PR #1, opened 2022-05-19
Implement an interface for the prerecorded API

yottalogical:


Cargo.toml
```diff
@@ -23,3 +24,4 @@ url = "2.2.2"
 [dev-dependencies]
 cpal = "0.13.5"
 crossbeam = "0.8.1"
+reqwest = { version = "0.11.10", features = ["stream"] }
```
andrewhalle, line 27: which example requires the `stream` feature? not 100% clear to me

src/lib.rs
```diff
@@ -24,6 +28,101 @@ where
     K: AsRef<str>,
 {
     api_key: K,
+    client: reqwest::Client,
+}
+
+#[non_exhaustive]
```
andrewhalle, line 34: Two things about this

1. I'm not sure it needs to be `#[non_exhaustive]`. I think non-exhaustive is primarily for structs we expect the consumer to match on, and we want to be able to add variants without breaking their code. This type will primarily be constructed by the consumer, not matched on (I think) so it's probably not necessary. Doesn't hurt anything though, I learned today by looking at [this](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) that the non-exhaustive attribute has no effect in the defining crate, so we still get the warnings about non-exhaustive patterns (which is good!)
2. The lower case variants look a little weird to my eye, I think they could be capitalized normally, since we have to turn them into strings anyway, so we can format them however we like here, but I don't mind that much

Cargo.toml
yottalogical: It's so that reqwest can read directly from a file. Without the `stream` feature, it would have to be done by reading whole file into memory first.

examples/prerecorded_from_file/main.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[tokio::main]
+async fn main() -> Result<(), DeepgramError> {
+    let dg_client = Deepgram::new(DEEPGRAM_API_KEY);
+
+    let file = File::open(PATH_TO_FILE).await.unwrap();
+
+    let source = BufferSource {
+        buffer: file,
+        mimetype: Mimetype::audio_wav,
```
andrewhalle, line 15: with our batch API, you don't have to specify the audio encoding (it'll be detected) so it may not make sense to require `mimetype` here

src/lib.rs
yottalogical: My idea was to have them be as similar to the actual mimetype names as possible, but I agree it does come at the cost of being less Rust-like. The same goes for the language codes below.

Cargo.toml
andrewhalle: gotcha, that makes sense

src/lib.rs
andrewhalle: Yep makes sense, totally your call, I don't mind very much any which way!

examples/prerecorded_from_file/main.rs
```diff
@@ -0,0 +1,26 @@
...
+    let source = BufferSource {
+        buffer: file,
+        mimetype: Some("audio/wav"),
+    };
+
+    let options = OptionsBuilder::new()
+        .punctuate(true)
+        .language(Language::en_US);
+
+    let response = dg_client.prerecorded_request(source, &options).await?;
```
andrewhalle, line 22: what do you like better API-wise between what you have here and
```
dg_client.prerecorded()
    .source(source)
    .punctuate(true)
    .language(Language::en_US)
    .send()
    .await?;
```

andrewhalle: I can see both being good right now, having a separate `Options` struct like what you have means that you can set that up once and then use it for multiple requests, which I really like. Putting it all on `PrerecordedRequestBuilder` means one less thing to import, which I thought I liked, but now that I'm looking at it I'm not sure

src/lib.rs
```diff
@@ -159,6 +261,128 @@ where
             channels: None,
         }
     }
+
+    pub async fn prerecorded_request(
+        &self,
+        source: impl AudioSource,
+        options: &OptionsBuilder<'_>,
```
andrewhalle, line 268: when I see types that end in `*Builder` they normally have a `.build()` method or similar which produces a `Options` type without the builder, you may want that `Options` type to be the parameter here, rather than the builder.

src/lib.rs
```diff
@@ -24,6 +29,93 @@ where
...
+}
+
+#[derive(Debug)]
+pub enum Utterances {
+    Disabled,
+    Enabled { utt_split: u32 },
+}
+
+#[derive(Debug)]
+pub struct OptionsBuilder<'a> {
```
andrewhalle, line 102: struct name could probably be a bit more specific

examples/prerecorded_from_file/main.rs
yottalogical: The way I have it currently is more similar to our other SDKs (the Python one, at least). I think I like your suggestion more, since it's a lot more similar to how Reqwest works, which Rust programmers might be more familiar with.

We'd probably want something more like the following, since the source isn't optional:

```rust
dg_client.prerecorded(source)
    .punctuate(true)
    .language(Language::en_US)
    .send()
    .await?;
```

examples/prerecorded_from_url.rs
```diff
@@ -0,0 +1,20 @@
+use deepgram::{Deepgram, DeepgramError, Language, OptionsBuilder, UrlSource};
+
+const DEEPGRAM_API_KEY: &str = "YOUR_SECRET";
```
jcdyer, line 3: Examples should be runnable by users who download the SDK.  For this, maybe consider using an environment variable instead of a const in the main function.

Side note:  For string literals, `static` is probably a more usual choice than `const`.  A const gets copied into every location it gets used, while a `static` gets assigned a location in the binary's read-only memory section.

src/prerecorded/options/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+    ko,
+    pt,
+    pr_BR,
+    ru,
+    es,
+    es_419,
+    sv,
+    tr,
+    uk,
+}
```
jcdyer, line 63: This is good for enumerating the languages we support, but now we need to keep this up to date whenever a new language gets added.  Is it worth adding an `Other(String)` variant, so we don't end up holding users back?  Or is that too confusing?  Maybe document it:

```rust
enum Language {
    ...
    tr,
    uk,
    /// We try to keep this list of languages up-to-date, but Deepgram moves fast and so do you.  
    /// If you need to transcribe languages that are supported by Deepgram, but aren't yet listed above, 
    /// you can supply a language code here.
    Other(String),
}
```

src/prerecorded/options/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+                Language::ko => "ko",
+                Language::pt => "pt",
+                Language::pr_BR => "pr_BR",
+                Language::ru => "ru",
+                Language::es => "es",
+                Language::es_419 => "es-419",
+                Language::sv => "sv",
+                Language::tr => "tr",
+                Language::uk => "uk",
+            };
```
jcdyer, line 259: Consider putting this whole match block in an `impl Display for Language`

src/prerecorded/options/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+        if let Some(profanity_filter) = profanity_filter {
+            seq.serialize_element(&("profanity_filter", profanity_filter))?;
+        }
+
+        for element in redact {
+            let s = match element {
+                Redact::Pci => "pci",
+                Redact::Numbers => "numbers",
+                Redact::Ssn => "ssn",
+            };
```
jcdyer, line 277: Also a candidate for `impl Display`

src/prerecorded/options/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+            Some(Utterances::Disabled) => seq.serialize_element(&("utterances", false))?,
+            Some(Utterances::Enabled { utt_split: None }) => {
+                seq.serialize_element(&("utterances", true))?
+            }
+            Some(Utterances::Enabled {
+                utt_split: Some(utt_split),
+            }) => {
+                seq.serialize_element(&("utterances", true))?;
+                seq.serialize_element(&("utt_split", utt_split))?;
+            }
```
jcdyer, line 324:
```suggestion
            Some(Utterances::Enabled { utt_split }) => {
                seq.serialize_element(&("utterances", true))?
                if let Some(utt_split) = utt_split {
                    seq.serialize_element(&("utt_split", utt_split))?;
                }
            }
```

src/prerecorded/options/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+        if let Some(tag) = tag {
+            seq.serialize_element(&("tag", tag))?;
+        }
+
+        seq.end()
+    }
+}
+
+#[cfg(test)]
+mod serialize_options_tests;
```
jcdyer, line 337: Normally, we keep unit test modules in-file with the module they test.
```rust
#[cfg(test)]
mod tests {
   ...
}
```

src/prerecorded/options/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+    tr,
+    uk,
+}
+
+#[derive(Debug)]
+pub enum Redact {
+    Pci,
+    Numbers,
+    Ssn,
+}
```
jcdyer, line 70: Not as worried about the closed set of redactions, as that doesn't change as frequently as supported languages.

jcdyer (commented):
I think some of these types will be used in both streaming and prerecorded, so it might make sense to pull some options into a separate module, not under either.  The code isn't public yet, so that could happen later, when it becomes more obvious where they make sense.

yottalogical:
The prerecorded part of the SDK is now operational. Still need to do enhancements, interface changes, tests, docs, bug fixes, etc.

examples/prerecorded_from_url.rs
yottalogical: Implemented in:
 * c24042cac528cb32091de10d69b4493a64983adb
 * 71200b650b6fb1ac6a0aa4d7af7c25ac13e65b85

src/prerecorded/options/mod.rs
yottalogical: Implemented in 63787b227a93fac0951f60dfdfe03c1bc81086b3.

src/prerecorded/options/mod.rs
yottalogical: Implemented in 07e3106542a8202f26d5e2a6fe7c4c190d897107.

src/prerecorded/options/mod.rs
yottalogical: I did try that at first, but `Display` is for building `String`s, not accessing `&str`s that already exist. It would work fine, but I didn't really want to add unnecessary overhead.

Is it worth doing anyways?

andrewhalle: generalizes nicely to the the `Other` case as well since it contains a `String`

yottalogical:
Similar to how we have an `Other` variant for the `Language` enum, do you think it would be a good idea to allow users to specify other things to put into the query string? That way they will be able to use new features added to Deepgram's API without being forced to update the SDK.

It might look something like this:

```rust
OptionsBuilder::new().custom_option("my_custom_option", "Waluigi");
```

andrewhalle:
I like that, it allows for the flexibility to add query params we don't actually support as well, which seems useless, but maybe they're accessing deepgram through a proxy on their side, and they want to build some custom functionality on top of that setup, controlled by query params.

An alternate idea is to allow getting the `Request` out of the options builder after it's built, for full flexibility

src/lib.rs
yottalogical: Implemented in 0d9cc647e793b93cc3f8526002ecde41d62785bd.

src/prerecorded/options/mod.rs
jcdyer: Sort of.  Display is for general formatting, which might involve creating a String, or might involve writing to a socket or a file using the `write!()` macro, which doesn't involve any allocation.

jcdyer: Also, I would prioritize simplicity (specifically, not putting lifetimes on structs) in setup code, since that is never the bottleneck for performance.  (Especially when sending hundreds of megabytes of audio over the wire.

jcdyer:
Similarly, I've seen good apis that allow bringing one's own Request object to the party, so you can set custom headers there.  I think tungstenite does something like that, though I wasn't very impressed with how they documented it, as there were some corner cases around headers getting overwritten or something.  Details are escaping me now.

src/prerecorded/options/mod.rs
andrewhalle: > specifically, not putting lifetimes on structs

I agree with that, probably better for the `Other` variant to hold a `String` instead of a `&str`

examples/make_prerecorded_request_builder.rs
```diff
@@ -0,0 +1,38 @@
...
+    let request_builder = dg_client.make_prerecorded_request_builder(&source, &options);
+
+    // Customize the RequestBuilder here
+    let customized_request_builder = request_builder
+        .query(&[("custom_query_key", "custom_query_value")])
+        .header("custom_header_key", "custom_header_value");
+
+    // It is necessary to annotate the type of response here
+    // That way it knows what type to deserialize the JSON into
+    let response: PrerecordedResponse = customized_request_builder.send().await?.json().await?;
```
yottalogical, line 32: Here's an example of it in action.

src/prerecorded/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+        let request_builder = self
+            .client
+            .post("https://api.deepgram.com/v1/listen")
+            .header("Authorization", format!("Token {}", self.api_key.as_ref()))
+            .query(&SerializableOptions(options));
+        let request_builder = source.fill_body(request_builder);
+
+        request_builder
+    }
+}
```
yottalogical, line 87: Here's the source and docs.

yottalogical (commented):
Okay, here's what I came up with in terms of customizing requests. It makes the same `reqwest::RequestBuilder` that we usually would, but gives it to the SDK user so they can further customize it before sending the request.

src/lib.rs
andrewhalle: maybe some variant of `PrerecordedOptionsBuilder` etc, not a strong preference

andrewhalle (commented):
I'm inclined to say let's merge this! I think there's some stuff that could be cleaned up/simplified, but this PR has grown large and I see no reason we can't do that cleanup in smaller, targeted, follow-up PRs. We don't have branch protection right now, and this isn't published to crates.io, so I say, if you're happy with it, merge it at your leisure!

yottalogical:
Some feedback on [the last commit](https://github.com/deepgram-devs/deepgram-rust-sdk/pull/1/commits/2cb04c7ee44ab0d2416e1fe3399e6a6afa3da81e) would be appreciated. It makes the way to invoke the [Callback feature](https://developers.deepgram.com/documentation/features/callback/) inconsistent with all the other features, but for (in my opinion) a good reason. See the commit message for more details.

examples/callback.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    let source = UrlSource { url: AUDIO_URL };
+
+    let options = Options::builder()
+        .punctuate(true)
+        .language(Language::en_US)
+        .build();
+
+    let response = dg_client
+        .callback_request(&source, &options, "https://example.com/")
```
andrewhalle, line 24: should we make the callback URL an environment variable, so that the example is runnable without editing it?

andrewhalle:
I agree that having a separate `callback_request` method makes sense to me

yottalogical:
I feel like it's ready to merge now. Last call for reviews!

src/prerecorded/audio_source.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// of the raw binary audio data.
+    pub mimetype: Option<&'a str>,
+}
+
+impl<'a, B: Borrow<UrlSource<'a>>> AudioSource for B {
+    fn fill_body(self, request_builder: RequestBuilder) -> RequestBuilder {
+        request_builder.json(self.borrow())
+    }
+}
+
```
andrewhalle, line 40: consider instead as a small simplification
```suggestion
impl<'a, 'b> AudioSource for &'a UrlSource<'b> {
    fn fill_body(self, request_builder: RequestBuilder) -> RequestBuilder {
        request_builder.json(self)
    }
}
```

src/prerecorded/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+
+static DEEPGRAM_API_URL_LISTEN: &str = "https://api.deepgram.com/v1/listen";
+
+impl<K: AsRef<str>> Deepgram<K> {
+    /// Sends a request to Deepgram to transcribe pre-recorded audio.
+    /// If you wish to use the Callback feature, you should use [`Deepgram::callback_request`] instead.
+    ///
+    /// The `source` parameter is either a [`BufferSource`] or a [`UrlSource`].
+    ///
+    /// See the [Deepgram API Reference](https://developers.deepgram.com/api-reference/#transcription-prerecorded) for more info.
```
andrewhalle, line 31: I'll leave this comment just here, but it applies to all the docs. The link is a little long, and it might format nicer if you extract it into a link reference, like
```
/// See the [Deepgram API Reference](api) for more info.
/// ...
/// [api]: https://developers.deepgram.com/api-reference/#transcription-prerecorded
```

src/prerecorded/options.rs
```diff
@@ -0,0 +1,1365 @@
...
+                models: Some(models.into_iter().collect()),
+            });
+        }
+
+        self
+    }
+
+    /// Set the maximum number of transcript alternatives to return.
+    ///
+    /// See the [Deepgram API Reference](https://developers.deepgram.com/api-reference/#alternatives-pr) for more info.
```
andrewhalle, line 509: you might have meant "Deepgram alternatives feature" here based on the surrounding docs

PR #2, opened 2022-06-06
Prepare to publish to crates.io

yottalogical:


Cargo.toml
```diff
@@ -1,26 +1,32 @@
...
+serde_json = "1.0.81"
+thiserror = "1.0.31"
+tokio = { version = "1.19.2", features = [ "full" ] }
 tokio-tungstenite = { version = "0.17.1", features = [ "native-tls" ] }
-tokio-util = { version = "0.7.1", features = [ "codec", "io" ] }
+tokio-util = { version = "0.7.3", features = [ "codec", "io" ] }
 tungstenite = "0.17.2"
 url = "2.2.2"
-uuid = { version = "1.0.0", features = ["serde"] }
+uuid = { version = "1.1.1", features = ["serde"] }
```
jcdyer, line 29: How strict do we want to  be about versions here?  It looks like we're specifying the current latest for everything.  This could make it harder to integrate the sdk with existing projects.

yottalogical: > It looks like we're specifying the current latest for everything.

That's what I did because I wasn't aware of the downsides, and wanted to make sure we're using the version with all the latest fixes and fixes.

Are there any packages that you think we should intentionally specify an older version of (or at least a way for me to figure out which)?

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,6 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Vonage Developer Community Slack
+    url: https://developer.nexmo.com/community/slack
+  - name: VonageDev on Twitter
+    url: https://twitter.com/VonageDev
```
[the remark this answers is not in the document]
yottalogical: @MichaelJolley, can you weigh in on this? I copied this file from [here](https://github.com/deepgram/oss-repo-template/blob/main/.github/ISSUE_TEMPLATE/config.yml).

Cargo.toml
yottalogical: Also, if this can't be figured out today, do you think it should hold back the release of version 0.1.0?

andrewhalle: In general, as I understand it, applications should specify the maximum version they can to make sure they have the most recent versions possible, but libraries should specify the minimum version they require so that more applications can make use of them. This is related to why we don't commit the `Cargo.lock` file for libraries, because our users may ultimately use different versions

Cargo.toml
```diff
@@ -1,26 +1,32 @@
 [package]
 name = "deepgram"
-version = "0.0.1"
+version = "0.1.0"
```
andrewhalle, line 3: I wouldn't mind keeping this at 0.0.1 to indicate that it's a very early version. Not that 0.1.0 is not a very early version as well, but I'd like to get a couple of iterations in before choosing 0.1.0. Willing to be overridden on this though @jcdyer

src/lib.rs
```diff
@@ -1,3 +1,7 @@
+#![forbid(unsafe_code)]
+#![warn(clippy::cargo)]
+#![allow(clippy::multiple_crate_versions)]
```
andrewhalle, line 3: what requires this `allow`?

Cargo.toml
yottalogical: The advantage of `0.1.0` is that it allows us to publish a [SemVer compatible update](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) (`0.1.1`) if/when necessary. I think this is why Cargo uses `0.1.0` as the default first version.

Cargo.toml
yottalogical: > so that more applications can make use of them

I've been thinking about this for a while, and I don't see how it applies. SemVer compatibility is supposed to be an [equivalence relation](https://en.wikipedia.org/wiki/Equivalence_relation).

Let's say H is the most recent version of some package P, and L is the lowest version that's compatible to H (example: H = 2.4.3, L = 2.0.0). If a binary crate (that depends on our crate) specifies a version of that package P, that version is compatible with L if and only if it's also compatible with H.

Now, not everyone is perfect with SemVer. Sometimes a "minor" update is published that accidentally includes a breaking change. But it seems like trying to specify a minimum version is also problematic.

When we compile and test our library, Cargo tries to use the most recent SemVer compatible version of every dependency. This means that although we might be promising to be compatible with `tokio` version 1.17.0, we actually might be using features introduced in 1.19.0 without realizing it, and the compiler isn't catching this for us. We could use nightly Rust to try and compile with `minimal-versions`, but that doesn't catch any problems we might have with version 1.18.0.

Specifying the versions the library has been tested against right before a publish seems like the best bet to me. But I'm also not an expert at dependency resolution, and might be weighing the factors wrong.

mlodato517: > We could use nightly Rust to try and compile with minimal-versions, but that doesn't catch any problems we might have with version 1.18.0.
> 
> Specifying the versions the library has been tested against right before a publish seems like the best bet to me. But I'm also not an expert at dependency resolution, and might be weighing the factors wrong.

I generally agree with this but I think if we don't attempt to trust the ecosystem we'll make tons of work for ourselves. I'd say to try to compile with minimum versions and deploy with that. If someone reports a bug with a particular version we can update the dependency to be the version above that one (depending on the root cause of their reported issue). That being said I'm probably just being lazy.

andrewhalle: It's possible for crates to specify a maximum dependency they allow (for example, to maintain a MSRV) so if a binary crate depends on us and another crate, and that crate has specified a maximum version of one of our dependencies that's not compatible, then that user is out of luck and has to pick between our crate and the other, even if we aren't using any of the features from the latest version of the dependency we specified, and could have built fine on the earlier version.

There's a procedure outlined in _Rust for Rustaceans_ for figuring out minimum dependency versions, which involves using `minimal-versions` and setting all our dependencies to their last major version, then incrementing minor versions until you get a successful build, then setting our dependency versions in `Cargo.toml` to whatever got picked in the `Cargo.lock`. Not saying we have to do that right now I think it's overkill for a first release, just relaying it as a general process we'll need to do at some point.

bowlofeggs: +1 for Cargo.toml specifying the minimum version that has the needed API.

.github/ISSUE_TEMPLATE/config.yml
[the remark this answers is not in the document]
MichaelJolley: Corrected.

Cargo.toml
yottalogical: Okay, I've pushed all the dependencies back to the oldest version that's still SemVer compatible with the latest version and works with our package. I also made the CI workflow run a build with `-Z minimal-versions` so we can always be sure it's working.

> I think it's overkill for a first release

Too bad. Did it anyways.

Cargo.toml
```diff
@@ -1,28 +1,34 @@
...
+url = "2.0.0"
 uuid = { version = "1.0.0", features = ["serde"] }
 
 [dev-dependencies]
-cpal = "0.13.5"
-crossbeam = "0.8.1"
-reqwest = { version = "0.11.10", features = ["stream"] }
+cpal = "0.13.0"
+crossbeam = "0.8.0"
+reqwest = { version = "0.11.7", features = ["stream"] }
```
jcdyer, line 34: This looks good now, assuming it still works with minimal version.  

Stylistic note: I like to use the minimum "significant digits" in library crates.  So if it works with any 1.y.z release say, serde, then I just put `serde = "1"`.  What you have here is semantically equivalent, and perfectly fine.

Cargo.toml
jcdyer: I start with 0.1.0, but have never given much thought to why.  I like @yottalogical 's reason here.

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,27 @@
+## Proposed changes
+
+Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.
+
+## Types of changes
+
+What types of changes does your code introduce to the Vonage for Visual Studio Code extension?
```
mlodato517, line 7: What's Vonage/Visual Studio Code? Is this copy-paste from some other template?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+jobs:
+  Build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install libasound2-dev
+      run: |
+        sudo apt-get update
+        sudo apt-get install libasound2-dev
```
mlodato517, line 17: Huh ... what's this for?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,71 @@
...
+  Documentation:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Cargo Doc
+      run: cargo doc --workspace --all-features
+  Audit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
```
mlodato517, line 55: What are the pros/cons to using https://github.com/actions-rs/audit-check?

mlodato517 (commented):
Didn't get through it all but I know you're on a tight timeline!

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
yottalogical: @MichaelJolley I think we caught another one.

.github/workflows/ci.yaml
yottalogical: The example code uses the [`cpal`](https://crates.io/crates/cpal) crate, which needs `libasound2-dev` to be installed in order to compile.

README.md
```diff
@@ -0,0 +1,45 @@
...
+To make sure our community is safe for all, be sure to review and agree to our
+[Code of Conduct](./CODE_OF_CONDUCT.md). Then see the
+[Contribution](./CONTRIBUTING.md) guidelines for more information.
+
+## Getting Help
+
+We love to hear from you so if you have questions, comments or find a bug in the
+project, let us know! You can either:
+
+- [Open an issue](https://github.com/deepgram/[reponame]/issues/new) on this repository
```
mlodato517, line 40: Was this copy-pasted from somewhere? I don't think we want `[reponame]` in there. [The link](https://github.com/deepgram/%5Breponame%5D/issues/new) doesn't seem to work through any GitHub magic.
```suggestion
- [Open an issue](https://github.com/deepgram-devs/deepgram-rust-sdk/issues/new) on this repository
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
...
+  * If you'd like to accomplish something in the extension that it doesn't already do, describe the problem in a new Github Issue.
+    * Issues that have been identified as a feature request will be labeled `enhancement`.
+    * If you'd like to implement the new feature, please wait for feedback from the project maintainers before spending
+      too much time writing the code. In some cases, `enhancement`s may not align well with the project objectives at
+      the time.
+* Tests, Documentation, Miscellaneous
+  * If you think the test coverage could be improved, the documentation could be clearer, you've got an alternative
+    implementation of something that may have more advantages, or any other change we would still be glad hear about
+    it.
+    * If its a trivial change, go ahead and send a Pull Request with the changes you have in mind
```
mlodato517, line 22:
```suggestion
    * If it's a trivial change, go ahead and send a Pull Request with the changes you have in mind
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
...
+    * If not, open a Github Issue to discuss the idea first.
+* Snippets
+  * To add snippets:
+    * Add a directory in the `snippets` folder with the name of the language.
+    * Add one or more files in the language directory with snippets.
+    * Update the `package.json` to include the snippets you added.
```
mlodato517, line 28: This might be copy-pasta?
```suggestion
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Requirements
+
+For a contribution to be accepted:
+
+* The test suite must be complete and pass
+* Code must follow existing styling conventions
+* Commit messages must be descriptive. Related issues should be mentioned by number.
+
+If the contribution doesn't meet these criteria, a maintainer will discuss it with you on the Issue. You can still
```
mlodato517, line 40: Would we discuss it on the Issue or on the PR? I'm guessing the PR but :shrug:

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+
+> What would you expect to happen when following the steps above?
+
+## Please tell us about your environment
+
+> We want to make sure the problem isn't specific to your operating system or programming language.
+  
+- **Operating System/Version:** Windows 10
+- **Language:** [all | TypeScript | Python | PHP | etc]
+- **Browser:** Chrome
```
mlodato517, line 28:
```suggestion
> We want to make sure the problem isn't specific to your operating system.
  
- **Operating System/Version:** Windows 10
- **Browser:** Chrome
```
I'd also vote to change Chrome to Firefox Just Because (cc @bowlofeggs)

mlodato517 (commented):
Found a broken link and some other probably-copy-:spaghetti:s

yottalogical:
It seems like there are quite a few issues with the [deepgram-devs/oss-repo-template](https://github.com/deepgram-devs/oss-repo-template).

.github/ISSUE_TEMPLATE/bug_report.md
bowlofeggs: Chrome is a pile of unmaintained vendored dependencies.

Cargo.toml
yottalogical: > There's a procedure outlined in _Rust for Rustaceans_ for figuring out minimum dependency versions

It's in Chapter 5 on page 82.

.github/workflows/ci.yaml
yottalogical: It's only a dev-dependency, so users of our crate won't need it.

mlodato517: Oh for sure. I was just hoping we could remove it from CI :-) But we can't since we `--all-targets` (which is good - we want those examples to keep compiling)

PR #3, opened 2022-06-08
Move reqwest stream feature to normal dependencies

yottalogical:
Previously it was just in the `dev-dependencies`, since it technically isn't needed to compile our library crate by itself.
However, it seems like it will be pretty common for people to want to send files, which requires the stream feature.

Users could enable the stream feature in their own dependencies, but since it's not obvious that that's the solution, it seems best to enable it here.


andrewhalle (approved):
seems reasonable!

PR #9, opened 2022-06-21
Define the remaining modules and clients for the API

yottalogical:
The organizational structure is identical to the [Deepgram API Reference](https://developers.deepgram.com/api-reference/).

Generating a specialized client from the `Deepgram` client is similar to how our existing SDKs work.

---

The following pull requests should be merged before this one:

 - #7


src/transcription.rs
```diff
@@ -18,13 +22,13 @@ pub struct Transcription<'a, K: AsRef<str>>(&'a Deepgram<K>);
 impl<'a, K: AsRef<str>> Deepgram<K> {
     /// Construct a new [`Transcription`] from a [`Deepgram`].
     pub fn transcription(&'a self) -> Transcription<'a, K> {
-        Transcription(self)
+        self.into()
     }
 }
```
jcdyer, line 27: Ah, interesting use of impl blocks outside the type's module.  Makes sense.

src/transcription/prerecorded.rs
```diff
@@ -81,11 +84,15 @@ impl<K: AsRef<str>> Transcription<'_, K> {
...
     ///
     /// ```no_run
+    /// # use std::env;
+    /// #
     /// # use deepgram::{
-    /// #     transcription::prerecorded::{Language, Options, UrlSource},
+    /// #     transcription::prerecorded::{
+    /// #         audio_source::UrlSource,
+    /// #         options::{Language, Options},
+    /// #     },
```
jcdyer, line 93: For discussion: Is it worth exposing these extra module layers to the users?  Might be better to make judicious use of `pub use` here, so the extra layers are only for organization convenience, but the API remains simple.

src/transcription/live.rs
```diff
@@ -2,7 +2,11 @@
 // Currently not documented because interface of this module is still changing
 #![allow(missing_docs)]
 
-//! Provides various types that are used for streaming transcription requests to Deepgram.
+//! Types used for live audio transcription.
+//!
+//! See the [Deepgram API Reference][api] for more info.
+//!
+//! [api]: https://developers.deepgram.com/api-reference/#transcription-streaming
```
jcdyer, line 9: If the documentation calls this mode "streaming" wouldn't it make sense to follow that same terminology here?  I don't see us using the term "live" other places.

yottalogical: I have been basing all naming so far off of the Python SDK. I do think that "streaming" makes more sense, though.

src/transcription/prerecorded.rs
yottalogical: This is something that I have given some thought to. I feel like the only layer worth skipping over would be the `transcription` layer.

Before:
```rust
use deepgram::transcription::prerecorded::{
    audio_source::UrlSource,
    options::{Language, Options},
};
```

After:
```rust
use deepgram::prerecorded::{
    audio_source::UrlSource,
    options::{Language, Options},
};
```

Beyond that I think the organization is useful. In version 0.1.0, I did have everything below the `prerecorded` layer re-exported, but that resulted in [this](https://docs.rs/deepgram/0.1.0/deepgram/prerecorded/index.html). There's a whole bunch of types here, and it's not immediately obvious what each one is used for.

PR #19, opened 2022-07-18
Add tier feature

yottalogical:


jcdyer:
> If a user wants to use the enhanced tier on one channel but not the other, is that even possible with the current API?

I don't think it's currently possible.

src/transcription/prerecorded/options.rs
```diff
@@ -939,6 +978,7 @@ impl Serialize for SerializableOptions<'_> {
 
         // Destructuring it makes sure that we don't forget to use any of it
```
jcdyer, line 979: That's a good trick. :+1:

src/transcription/prerecorded/options.rs
```diff
@@ -1235,7 +1290,17 @@ mod serialize_options_tests {
             .tag(["Tag 1"])
             .build();
 
-        check_serialization(&options, "model=finance%3Aextra_crispy%3Aconversationalai&version=1.2.3&language=en-US&punctuate=true&profanity_filter=true&redact=pci&redact=ssn&diarize=true&ner=true&multichannel=true&alternatives=4&numerals=true&search=Rust&search=Deepgram&keywords=Ferris&keywords=Cargo%3A-1.5&utterances=true&utt_split=0.9&tag=Tag+1");
+        check_serialization(&options, "tier=enhanced&model=finance%3Aextra_crispy%3Aconversationalai&version=1.2.3&language=en-US&punctuate=true&profanity_filter=true&redact=pci&redact=ssn&diarize=true&ner=true&multichannel=true&alternatives=4&numerals=true&search=Rust&search=Deepgram&keywords=Ferris&keywords=Cargo%3A-1.5&utterances=true&utt_split=0.9&tag=Tag+1");
```
jcdyer, line 1293: `extra_crispy`?

jcdyer (approved):

jcdyer:
Apparently my approval is meaningless here.

src/transcription/prerecorded/options.rs
yottalogical: For when your normal transcripts just don't have enough crunch to them.

PR #21, opened 2022-08-03
Make AudioSource a type rather than a trait

yottalogical:
This change makes it easier to work with the SDK if it isn't known until runtime whether you'll be using a URL source or a buffer source. This was based off of a recommendation from @ChristopherBiscardi in [this video](https://youtu.be/abdEyc9LyhU) from the experience of building [Lamarck](https://github.com/ChristopherBiscardi/lamarck).

Originally I had avoided using an enum because putting generics on enums can be a nightmare for the user of the library, but @jcdyer pointed out a way to do it without generics entirely.


src/transcription/prerecorded/audio_source.rs
```diff
@@ -2,67 +2,69 @@ use reqwest::{header::CONTENT_TYPE, RequestBuilder};
...
-impl AudioSource for &UrlSource<'_> {
-    fn fill_body(self, request_builder: RequestBuilder) -> RequestBuilder {
-        request_builder.json(self)
+    /// Same as [`AudioSource::from_buffer`], but allows you to specify a [MIME type][mime].
+    ///
+    /// [mime]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#audio_and_video_types
+    pub fn from_buffer_with_mime_type(buffer: impl Into<reqwest::Body>, mime_type: &str) -> Self {
+        Self(InternalAudioSource::Buffer {
+            buffer: buffer.into(),
+            mime_type: Some(String::from(mime_type)),
```
jcdyer, line 45: Open for discussion: Is it worth using the [`mime::Mime`](https://docs.rs/mime/0.3.16/mime/struct.Mime.html) type for this?

src/transcription/prerecorded/audio_source.rs
```diff
@@ -2,67 +2,69 @@ use reqwest::{header::CONTENT_TYPE, RequestBuilder};
...
+#[derive(Debug)]
+pub struct AudioSource(InternalAudioSource);
 
-    pub trait Sealed {}
-
-    impl Sealed for &UrlSource<'_> {}
-    impl<B: Into<reqwest::Body>> Sealed for BufferSource<'_, B> {}
+#[derive(Debug)]
+enum InternalAudioSource {
+    Url(String),
```
jcdyer, line 10: Similarly, this could be the [`url::Url`](https://docs.rs/url/latest/url/struct.Url.html) type, though I'm less sold on that, since it doesn't remove the need for creating a string in any case.

src/transcription/prerecorded/audio_source.rs
jcdyer: docs could provide an example like
```rust
let source = AudioSource::from_buffer_with_mimetype(buffer, "audio/mp3".parse()?);
````

File: src/transcription/prerecorded/audio_source.rs [hunk 3]
```diff
@@ -2,67 +2,69 @@ use reqwest::{header::CONTENT_TYPE, RequestBuilder};
...
-#[derive(Debug, PartialEq, Eq, Clone, Copy, Hash, Serialize)]
-pub struct UrlSource<'a> {
-    /// URL of audio file to transcribe.
-    pub url: &'a str,
-}
+impl AudioSource {
+    /// Constructs an [`AudioSource`] that will instruct Deepgram to download the audio from the specified URL.
+    pub fn from_url(url: &str) -> Self {
+        Self(InternalAudioSource::Url(String::from(url)))
+    }
```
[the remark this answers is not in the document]
yottalogical: Or how about `impl Into<String>`? That way they can do either a `String` or a `&str`, but it won't result in a double allocation if they pass a `String`.

I know that wrapping a `&str` literal in `String::from` isn't difficult, but it's a minor annoyance that I'd rather not force the user into.

src/transcription/prerecorded/audio_source.rs
yottalogical: I think this would make sense if the majority of users coming to this part already had a `Mime` object or if we were doing something with it other than just turning it back into a string.

It doesn't eliminate the possibility of them getting the MIME type wrong (for example, `"auido/wav"` parses without error). It would just catch some typos before the network request rather than after.

I don't feel like this is worth making using the SDK more complicated.

src/transcription/prerecorded/audio_source.rs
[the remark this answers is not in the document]
yottalogical: Implemented in https://github.com/deepgram-devs/deepgram-rust-sdk/pull/21/commits/edbc59bf32dfac49b92f0f51f2646d41930174e1.

PR #28, opened 2022-09-01
Derive `Serialize` for all response types

### LINKED ISSUE #27: Serialize for Response
## Proposed changes

It would be nice if the `Response` had a Serialize implementation so that it could be serialized/deserialized for tests.

## Context

The alternative is burning credits on the same transcription over and over when building on top of the SDK.

It would be nice to serialize a real-world response to a json file and then read that in for tests instead of needing to run a transcription again.

## Possible Implementation

Add a Serialize derive here: https://github.com/deepgram-devs/deepgram-rust-sdk/blob/daa8912fc26fd09b9a37984594b66aceae0d401e/src/transcription/prerecorded/response.rs#L15

andrewhalle:
**Why?**
Fixes #27.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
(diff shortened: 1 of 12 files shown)


andrewhalle:
Ah yeah, fair enough. I'll put an `allow` for this lint for now and leave it to be decided later.

PR #50, opened 2023-08-04
refactor!: remove generic from `Deepgram`

### LINKED ISSUE #49: Ungenericify Deepgram struct
## Proposed changes

The `Deepgram<K>` struct is generic over the API key `K`. The signature is:

```rust
pub struct Deepgram<K>
where
    K: AsRef<str>,
```

I propose `K` is cloned and stored internally in a `String`.

## Context

There is no obvious benefit to the API user to be generic over an API key. I can see two reasons to do this, and neither seem to hold up.

  1. Save space. However it's 40 bytes as a string (and looks like a hexadecimal number, so I assume it can be stored as a number too?). It seems to be a very small gain.
  2. Allow user to control storage of the API key. In a security paranoid environment maybe it would be possible to hold the key in a "secure enclave" or similar tech. Though `AsRef` would defeat this, since a simple operation is supposed to release the key. If I don't want the `Deepgram` struct to remain in memory with the key, I can simply drop the struct. If the API want to ensure the key is zeroed out from memory, the API can `impl Drop for Deepgram` and zero the memory where the key was stored.

## Possible Implementation

```rust
 fn new(key: impl AsRef<str>) -> Self {
    let api_key = key.as_ref().to_owned();
    Self { api_key }
 }
```

andrewhalle:
The `Deepgram` struct is unnecessarily generic. Storing the API key as a generic offers flexibility which probably isn't warranted by any use cases.

Fixes #49

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) doc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+- Remove generic from `Deepgram` struct.
+
 ## [0.3.0]
 
 ### Added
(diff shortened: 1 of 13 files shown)


mweber15 (approved):
Looks good to me!

PR #53, opened 2023-08-16
Proposal: make tokio optional

cardoso:
It should be possible to use the library without picking a runtime.

The live feature would take some work to uncouple from tokio and tungstenite, so I left it under a default flag for now.

### DIFF
--- a/src/transcription/live.rs
+++ b/src/transcription/live.rs
@@ -1,3 +1,4 @@
+#![cfg(feature = "tokio")]
 // TODO: Remove this lint
 // Currently not documented because interface of this module is still changing
 #![allow(missing_docs)]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,10 @@ repository = "https://github.com/deepgram-devs/deepgram-rust-sdk"
 keywords = ["deepgram", "asr", "transcription", "ai", "speech-to-text"]
 categories = ["api-bindings", "multimedia::audio"]
 
+[features]
+default = ["live"]
+live = ["tokio", "tokio-tungstenite", "tokio-util"]
+
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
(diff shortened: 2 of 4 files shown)


andrewhalle:
Thanks for the PR!

In principle, I'm okay with the idea of being able to use any runtime. However, our dependency on `tokio` runs a little deeper than the explicit dependency. `reqwest` only works with `tokio`, and we use `reqwest` to make the HTTP request.

Are you using `async-std`?

cardoso:
@andrewhalle Im using smol and it's possible to use reqwest with a different runtime using the async-compat crate: https://github.com/smol-rs/async-compat

andrewhalle:
@cardoso Ah okay, I understand now, thanks! Just to confirm, you're compiling for a WASM target? Otherwise, `reqwest` is still pulling in the `tokio` dependency.

andrewhalle:
I see that `async-compat` has a `tokio` dependency as well, so even if we make `tokio` optional in this crate, you'll still be getting it from that, right?

cardoso:
@andrewhalle yes, but not the runtime, so it's good enough for now to make the SDK usable with other runtimes and in wasm targets.

It would be good to make reqwest optional as well in the future, so an async runtime wouldn't be needed to access the pre-recorded API, but I wanted to make this PR short.

andrewhalle:
Cool, sounds good then. I'll give this a review, and we'll get this in. Thanks!

src/transcription/live.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg(feature = "tokio")]
```
andrewhalle, line 1:
```suggestion
#![cfg(feature = "live")]
```

andrewhalle:
@cardoso Just checking back in on this, is this still something you're interested in merging?

Boscop:
@cardoso Any update on this?
It would be great if we could use this crate on WASM 🙂

cardoso:
Sorry I don't have time to continue this PR. I suggest someone take over.

PR #68, opened 2024-07-03
Featurize transcription modes

jcdyer:
## Proposed changes

Add feature flags for `live` and `prerecorded` ASR.  This helps non-streaming users not have to import tungstenite.  I'm open to not doing things this way.  I'm thinking of also following up with a little more specificity about which tokio features to include for similar reasons.

## Types of changes

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,9 +22,9 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 thiserror = "1"
 tokio = { version = "1", features = ["full"] }
-tokio-tungstenite = { version = "0.20.1", features = ["rustls-tls-webpki-roots"] }
+tokio-tungstenite = { version = "0.20.1", features = ["rustls-tls-webpki-roots"], optional = true }
 tokio-util = { version = "0.7.1", features = ["codec", "io"] }
-tungstenite = "0.20.1"
+tungstenite = { version = "0.20.1", optional = true }
 url = "2"
 uuid = { version = "1", features = ["serde"] }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -34,3 +34,8 @@ proc-macro2 = "1.0.60"
 pkg-config = "0.3.27"
 cpal = "0.13"
 crossbeam = "0.8"
+
+[features]
+default = ["prerecorded", "live"]
+live = ["dep:tungstenite", "dep:tokio-tungstenite"]
```
andrewhalle, line 40: Use of `dep:` bumps our MSRV to 1.60. I don't think we document an MSRV, but we should and it'll need to be at least 1.60.

andrewhalle (approved):
I swear we had these feature flags already, they must have been removed at some point in the past

Cargo.toml
jcdyer: Whoa.  rust 1.60.0 is from April 2022.

jcdyer: I see the feature flags in a PR that never got merged.  Maybe that's what you were thinking of.

andrewhalle: That must be it

jcdyer:
Oops.  This PR breaks the examples.  (Found doing `cargo test --no-default-features`.  I guess I'll need to annotate the required features for each one.

jcdyer:
Also updated CI to run `check` with no features and with each feature in isolation.

andrewhalle (approved):

PR #70, opened 2024-07-03
Custom hosts

jcdyer:
## Proposed changes

Support connecting to deepgram hosts at URLs other than `https://api.deepgram.com`.  This will help us support self-hosted users, as well as supporting in-house development.

## Types of changes

## Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,11 +35,8 @@ static DEEPGRAM_BASE_URL: &str = "https://api.deepgram.com";
 #[derive(Debug, Clone)]
 pub struct Deepgram {
     #[cfg_attr(not(feature = "live"), allow(unused))]
-    api_key: String,
-    #[cfg_attr(
-        all(not(feature = "live"), not(feature = "prerecorded")),
-        allow(unused)
-    )]
+    api_key: Option<String>,
+    #[cfg_attr(not(any(feature = "live", feature = "prerecorded")), allow(unused))]
     base_url: Url,
     client: reqwest::Client,
 }
@@ -89,7 +86,7 @@ type Result<T> = std::result::Result<T, DeepgramError>;
 impl Deepgram {
     /// Construct a new Deepgram client.
     ///
-    /// The client will be pointed at the deepgram API.
+    /// The client will be pointed at Deepgram's hosted API.
     ///
     /// Create your first API key on the [Deepgram Console][console].
     ///
@@ -99,7 +96,8 @@ impl Deepgram {
     ///
     /// Panics under the same conditions as [`reqwest::Client::new`].
     pub fn new<K: AsRef<str>>(api_key: K) -> Self {
-        Self::with_base_url(api_key, DEEPGRAM_BASE_URL)
+        let api_key = Some(api_key.as_ref().to_owned());
+        Self::inner_constructor(DEEPGRAM_BASE_URL.try_into().unwrap(), api_key)
     }
 
     /// Construct a new Deepgram client with the specified base URL.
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -26,13 +27,20 @@ pub mod usage;
...
+
 /// A client for the Deepgram API.
 ///
 /// Make transcriptions requests using [`Deepgram::transcription`].
 #[derive(Debug, Clone)]
 pub struct Deepgram {
     #[cfg_attr(not(feature = "live"), allow(unused))]
     api_key: String,
+    #[cfg_attr(
+        all(not(feature = "live"), not(feature = "prerecorded")),
```
jgberry, line 40: Nit: I'd find this a bit easier to read as `not(any(feature = "live", feature = "prerecorded"))`.

src/lib.rs
```diff
@@ -81,6 +89,8 @@ type Result<T> = std::result::Result<T, DeepgramError>;
 impl Deepgram {
     /// Construct a new Deepgram client.
     ///
+    /// The client will be pointed at the deepgram API.
```
jgberry, line 92: Perhaps this should read:
```suggestion
    /// The client will be pointed at the hosted deepgram API.
```

src/lib.rs
```diff
@@ -89,6 +99,46 @@ impl Deepgram {
...
+    /// );
+    /// ```
+    ///
+    /// # Panics
+    ///
+    /// Panics under the same conditions as [`reqwest::Client::new`], or if `base_url`
+    /// is not a valid URL.
+    ///
+    ///
+
```
jgberry, line 135: Is there a reason for all this blank space here?

src/lib.rs
```diff
@@ -89,6 +99,46 @@ impl Deepgram {
...
+    pub fn with_base_url<K, U>(api_key: K, base_url: U) -> Self
```
jgberry, line 136: I find the signature of this function confusing for use with self-hosted. It seems to imply that self-hosted requests require an API key, when that is not the case. Even more concerning, if users did provide an API key, we would unnecessarily be passing that API key onto the self-hosted instance. We probably don't want to be doing that.

src/lib.rs
```diff
@@ -26,13 +27,20 @@ pub mod usage;
...
 
+static DEEPGRAM_BASE_URL: &str = "https://api.deepgram.com";
+
 /// A client for the Deepgram API.
 ///
 /// Make transcriptions requests using [`Deepgram::transcription`].
 #[derive(Debug, Clone)]
 pub struct Deepgram {
     #[cfg_attr(not(feature = "live"), allow(unused))]
     api_key: String,
```
jgberry, line 38: Not the fault of this PR, but the `api_key` is leaked in the `Debug` impl for this struct.

jcdyer: Good catch.

src/lib.rs
jcdyer: That makes sense.  I'll make it optional here, and explain the asymmetry between the signatures of `with_base_url` and `new` in the docs.

jcdyer: So having an `Option<impl AsRef<str>>` for `with_base_url` was confusing, because it meant that omitting it would require you to write `None::<String>` (or some other type that matches the generic constraint.  Instead I'm making two functions. `with_base_url` and `with_base_url_and_api_key`.  I'm keeping the api_key-only method named `new` since most users will just use that interface.  I'm almost at the point of wanting a builder pattern here.

src/lib.rs
jcdyer: I opted for "Deepgram's hosted API".  I think there's slightly more opportunity for confusion between "hosted Deepgram API" and "self-hosted Deepgram API".  (It's a Deepgram API, but who's doing the hosting?) "Deepgram's hosted API" directly answers the question of who's doing the hosting.

src/lib.rs
jcdyer: Nope.

src/lib.rs
jcdyer: Done.

jgberry (approved):

PR #79, opened 2024-07-25
WIP - Additional features

DamienDeepgram:
### Websocket Messages

- CloseStream
- Finalize

### Websocket Lifecycle Events

- Open
- Close
- Error
- Result

## Types of changes

## Checklist


examples/transcription/websocket/microphone_stream.rs
```diff
@@ -92,21 +102,55 @@ fn microphone_as_stream() -> FuturesReceiver<Result<Bytes, RecvError>> {
...
+            connection.finalize(event_tx.clone()).await?;
+
+            // Call  after processing the stream
+            connection.finish(event_tx.clone()).await?;
+        }
+        count += 1;
+        match response {
+            Ok(result) => println!("Transcription result: {:?}", result),
+            Err(e) => eprintln!("Transcription error: {:?}", e),
+        }
```
jcdyer, line 153: This is a weird mix of high and low level interface.  If the user provides a stream, all the details of managing the stream should Just Work :tm: .  They get back a stream of responses, and when the stream is exhausted (or an error occurs), the stream closes itself before returning None.

In 0.6.1 we can add an interface where they can also be request a LiveTranscriptionHandle (naming open for debate) instead of calling `.stream()`, at which point, they drive the whole thing: sending chunks of audio, receiving responses, finalizing the request, etc.

jcdyer: In other words, there should be separate interfaces or getting back a `response_stream` and a `connection`, instead of returning both at the same time.  The proposed interface does not make it clear what the user's responsibilities are vs. what the library handles.

src/lib.rs
```diff
@@ -143,14 +143,24 @@ pub enum DeepgramError {
...
+
+    /// Something went wrong with sending
+    #[cfg(feature = "listen")]
+    #[error("Something went wrong with WS: {0}")]
+    SendError(#[from] futures::channel::mpsc::SendError),
+
+    /// Something went wrong with receiving
+    #[cfg(feature = "listen")]
+    #[error("Deepgram error: {0}")]
+    CustomError(String),
```
jcdyer, line 163: String errors are generally a weird design.  Instead, let's return a dynamically dispatched Error, `Box<dyn Error>`, which means anything that implements the `Error` trait, unified to a single type.

src/listen/websocket.rs
```diff
@@ -333,26 +409,35 @@ where
                         let mut write = write_clone.lock().await;
                         if let Err(e) = write.send(keep_alive_message).await {
                             eprintln!("Error Sending Keep Alive: {:?}", e);
+                            let _ = event_tx_keep_alive
+                                .send(Event::Error(DeepgramError::from(e)))
+                                .await;
                             break;
```
jcdyer, line 415: TBD (in a later PR): we should only send keepalives when we haven't sent any other messages in a given time frame.

src/listen/websocket.rs
```diff
@@ -238,12 +257,60 @@ fn options_to_query_string(options: &Options) -> String {
...
+pub struct TranscriptionStream {
+    write_arc: Arc<Mutex<SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, Message>>>,
+}
+
+impl TranscriptionStream {
+    pub async fn finalize(
+        &self,
+        event_tx: Sender<Event>,
+    ) -> std::result::Result<(), DeepgramError> {
+        let finalize_message = Message::Text("{\"type\": \"Finalize\"}".to_string());
```
jcdyer, line 270: Idiomatic rust suggestion: use raw strings to avoid escaping quote characters.  A rust raw string consists of an `r` followed by zero or more `#`s (usually one) followed by a double quote, then the string, then the closing quote, and finally the same number of `#`s as at the start of the string.  Allowing any number of `#`s means you can include strings that contain `"##` sequences of any length.  So for example:
```rust
let string = r###"This is a r##"raw string that contains a r#"raw "string""# and a r"raw string with a \n" so when it ends here"## the text is still part of one big string"###;

But more simply, and more normally:

```suggestion
        let finalize_message = Message::Text(r#"{"type": "Finalize"}"#.to_string());
```

src/listen/websocket.rs
```diff
@@ -238,12 +257,60 @@ fn options_to_query_string(options: &Options) -> String {
...
+            event_tx.send(Event::Error(err)).await.unwrap();
+            return Err(DeepgramError::CustomError(
+                "Failed to send Finalize message".to_string(),
+            ));
+        }
+        Ok(())
+    }
+
+    pub async fn finish(&self, event_tx: Sender<Event>) -> std::result::Result<(), DeepgramError> {
+        let finish_message = Message::Text("{\"type\": \"CloseStream\"}".to_string());
```
jcdyer, line 284:
```suggestion
        let finish_message = Message::Text(r#"{"type": "CloseStream"}"#.to_string());
```

examples/transcription/websocket/microphone_stream.rs
```diff
@@ -92,21 +102,55 @@ fn microphone_as_stream() -> FuturesReceiver<Result<Bytes, RecvError>> {
...
+    let options = Options::builder()
+        .model(Model::Nova2)
+        .smart_format(true)
+        .language(Language::en_US)
+        .build();
+
+    let (connection, mut response_stream) = dg
         .transcription()
-        .stream_request()
+        .stream_request_with_options(Some(&options))
```
jcdyer, line 127:
```suggestion
        .stream_request_with_options(options)
```

src/listen/websocket.rs
```diff
@@ -238,12 +257,60 @@ fn options_to_query_string(options: &Options) -> String {
     serde_urlencoded::to_string(serialized_options).unwrap_or_default()
 }
 
+#[derive(Debug)]
+pub struct TranscriptionStream {
```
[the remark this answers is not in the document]
jcdyer: We should also `derive(Clone)` for it, since it just contains an `Arc`ed stream.

src/listen/websocket.rs
```diff
@@ -238,12 +257,60 @@ fn options_to_query_string(options: &Options) -> String {
...
+    write_arc: Arc<Mutex<SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, Message>>>,
+}
+
+impl TranscriptionStream {
+    pub async fn finalize(
+        &self,
+        event_tx: Sender<Event>,
```
jcdyer, line 268: `TranscriptionStream` should hold on to a `Sender<Event>` so the user doesn't have to pass it in to every message.

src/listen/websocket.rs
```diff
@@ -238,12 +257,60 @@ fn options_to_query_string(options: &Options) -> String {
...
+            eprintln!("Error sending Finalize message: {:?}", err);
+            event_tx.send(Event::Error(err)).await.unwrap();
+            return Err(DeepgramError::CustomError(
+                "Failed to send Finalize message".to_string(),
+            ));
+        }
+        Ok(())
+    }
+
+    pub async fn finish(&self, event_tx: Sender<Event>) -> std::result::Result<(), DeepgramError> {
```
jcdyer, line 283: Let's call this `close()` instead of `finish()`.  Finish is too similar to `finalize()` for my tastes, and doesn't match our docs.

src/listen/websocket.rs
```diff
@@ -317,13 +384,22 @@ where
...
-        let write = Arc::new(Mutex::new(write));
+        let write_arc = Arc::new(Mutex::new(write));
+        let cloned_write_arc = Arc::clone(&write_arc);
+        self.write_arc = Some(cloned_write_arc); // Store the write arc
         let (mut tx, rx) = mpsc::channel::<Result<StreamResponse>>(1);
 
+        let event_tx_open = event_tx.clone();
+        let event_tx_keep_alive = event_tx.clone();
+        let event_tx_send = event_tx.clone();
+        let event_tx_receive = event_tx.clone();
```
jcdyer, line 395: Nit: This is fine, but I would create these clones when you need them, and just call each one `event_tx`, using the same shadowing pattern as in the `write_arc` comments above.

src/listen/websocket.rs
```diff
@@ -361,23 +446,48 @@ where
             let mut write = write_clone.lock().await;
             if let Err(e) = write.send(Message::binary([])).await {
```
jcdyer, line 447: We should send a proper close message here, so the user doesn't have to explicitly interact with the stream to get proper closing behavior when the stream is done (this gets rid of the need for interacting with a `TranscriptionStream` while using the stream-driven interface.

jcdyer (changes requested):
This will need to be rebased onto the latest main branch to make the suggested changes.

The main request here is that we completely separate the stream handling interface from the (lower-level) manual message passing interface.  In the streaming case, you'll have something like (untested code);

```rust
    let (response_tx, response_rx) = futures_mpsc::channel(100);
    dg.transcription()
        .stream_request()
        .keep_alive()
        .stream(microphone_as_stream())
        .encoding(Encoding::Linear16)
        .sample_rate(44100)
        .channels(1)
        .start(response_tx)?;
    while let Some(event) = response_rx.next().await {
        println!("received event: {event:?}");
    }
    println!("The stream has closed and the websocket was closed cleanly");
```

In the Handle case, it would look like:
```rust
    let (response_tx, response_rx) = futures_mpsc::channel(100);
    let mut handle = dg.transcription()
        .stream_request()
        .keep_alive()
        .stream(microphone_as_stream())
        .encoding(Encoding::Linear16)
        .sample_rate(44100)
        .channels(1)
        .handle(response_tx)?;
    let mut audio = microphone_as_stream();
    let mut stream_done = false;
    loop {
        select! {
            event = response_rx.next() => {
                match event {
                    Some(event) => println!("Received an event: {event:?}),
                    None => break, // stream is closed
            audio = audio.next() => {
                match audio {
                    Some(audio) => handle.send_audio(audio).await?,
                    None => {
                        if !stream_done {
                            handle.finalize().await?;
                            handle.close().await?;
                            stream_done = true
                        }
                    }
                }
            }
        }
    }
    println!("The stream has closed and the websocket was closed cleanly");
```

src/listen/websocket.rs
```diff
@@ -238,12 +257,60 @@ fn options_to_query_string(options: &Options) -> String {
...
     E: Send + std::fmt::Debug,
 {
-    pub async fn start(self) -> Result<Receiver<Result<StreamResponse>>> {
+    pub async fn start(
+        mut self,
+        event_tx: Sender<Event>,
+    ) -> std::result::Result<
+        (
+            TranscriptionStream,
+            futures_mpsc::Receiver<std::result::Result<StreamResponse, DeepgramError>>,
```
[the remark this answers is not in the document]
jcdyer: Actually, hang on.  I'm unclear on something: what's the difference between the `Sender<Event>` and the returned `Receiver<Result<StreamResponse>>`?  Does the user need to listen to both of those?  I see an `Event::Result` type, but it just holds a `String`, and it's (maybe) never written to?

src/listen/websocket.rs
```diff
@@ -238,12 +257,60 @@ fn options_to_query_string(options: &Options) -> String {
     serde_urlencoded::to_string(serialized_options).unwrap_or_default()
 }
 
+#[derive(Debug)]
+pub struct TranscriptionStream {
+    write_arc: Arc<Mutex<SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, Message>>>,
+}
+
+impl TranscriptionStream {
```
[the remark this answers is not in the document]
DamienDeepgram: I don't think our other SDKs support ping/pong and not sure the extent of support on the server side either

src/listen/websocket.rs
DamienDeepgram: This matches the other SDKs.

See: https://github.com/deepgram/deepgram-python-sdk/blob/main/examples/speech-to-text/websocket/microphone/main.py#L134

This is not actually a close of the websocket it tells the websocket to finish processing all it's audio and then the server closes

src/listen/websocket.rs
DamienDeepgram: The server will close the connection if it never receives any audio even if you send keep alives

src/listen/websocket.rs
DamienDeepgram: Resolved

src/listen/websocket.rs
DamienDeepgram: Resolved

examples/transcription/websocket/microphone_stream.rs
DamienDeepgram: Resolved

src/listen/websocket.rs
```diff
@@ -317,13 +384,22 @@ where
...
+        let event_tx_send = event_tx.clone();
+        let event_tx_receive = event_tx.clone();
+
+        event_tx_open.send(Event::Open).await.unwrap();
+
         // Spawn the keep-alive task
         if self.keep_alive.unwrap_or(false) {
             {
-                let write_clone = Arc::clone(&write);
+                let write_clone = Arc::clone(&write_arc);
```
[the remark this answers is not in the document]
DamienDeepgram: I get an error if I do this

```
mismatched types
expected reference `&Arc<_, _>`
   found reference `&SplitSink<WebSocketStream<tokio_tungstenite::MaybeTlsStream<tokio::net::TcpStream>>, tungstenite::Message>`
```

src/listen/websocket.rs
```diff
@@ -333,26 +409,35 @@ where
...
         let send_task = async move {
             while let Some(frame) = source.next().await {
                 match frame {
                     Ok(frame) => {
                         let mut write = write_clone.lock().await;
                         if let Err(e) = write.send(frame).await {
                             println!("Error sending frame: {:?}", e);
+                            let _ = event_tx_send
+                                .send(Event::Error(DeepgramError::from(e)))
+                                .await;
```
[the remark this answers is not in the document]
DamienDeepgram: Resolved

src/listen/websocket.rs
```diff
@@ -317,13 +384,22 @@ where
         };
         let (ws_stream, _) = tokio_tungstenite::connect_async(request).await?;
         let (write, mut read) = ws_stream.split();
-        let write = Arc::new(Mutex::new(write));
+        let write_arc = Arc::new(Mutex::new(write));
+        let cloned_write_arc = Arc::clone(&write_arc);
+        self.write_arc = Some(cloned_write_arc); // Store the write arc
         let (mut tx, rx) = mpsc::channel::<Result<StreamResponse>>(1);
 
+        let event_tx_open = event_tx.clone();
```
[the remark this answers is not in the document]
DamienDeepgram: Resolved

src/listen/websocket.rs
```diff
@@ -361,23 +446,48 @@ where
...
+                                println!("Received close frame: {:?}", close_frame);
+                                // Send a close frame back to acknowledge the close request
+                                let mut write = recv_write_clone.lock().await;
+                                if let Err(_e) = write.send(Message::Close(None)).await {
+                                    // Should we notify the client we could not ack the server initiated close frame?
+                                    // eprintln!("Failed to send close frame: {:?}", e);
+                                    // let _ = event_tx_receive
+                                    //     .send(Event::Error(DeepgramError::from(e)))
+                                    //     .await;
+                                }
```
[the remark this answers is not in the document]
DamienDeepgram: Resolved but not sure how to type StreamCloseError

src/listen/websocket.rs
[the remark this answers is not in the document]
DamienDeepgram: Resolved

src/listen/websocket.rs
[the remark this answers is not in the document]
DamienDeepgram: The transcript comes back in the Receiver<Result<StreamResponse>>

eg

```
while let Some(response) = response_stream.next().await {
        match response {
            Ok(result) => println!("Transcription result: {:?}", result),
            Err(e) => eprintln!("Transcription error: {:?}", e),
        }
    }
```

src/listen/websocket.rs
DamienDeepgram: Resolved

src/listen/websocket.rs
DamienDeepgram: Not sure I fully understood this but moved them to where they get used and only create the keep alive one if needed

src/listen/websocket.rs
DamienDeepgram: Resolved *I think

DamienDeepgram:
Closing in favor of the different approach already merged

PR #80, opened 2024-07-25
Prep work for adding websocket handle

jcdyer:
## Types of changes

## Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,7 +154,7 @@ pub enum DeepgramError {
 }
 
 #[cfg_attr(not(feature = "listen"), allow(unused))]
-type Result<T> = std::result::Result<T, DeepgramError>;
+type Result<T, E=DeepgramError> = std::result::Result<T, E>;
 
 impl Deepgram {
     /// Construct a new Deepgram client.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,7 +154,7 @@ pub enum DeepgramError {
 }
 
 #[cfg_attr(not(feature = "listen"), allow(unused))]
-type Result<T, E=DeepgramError> = std::result::Result<T, E>;
+type Result<T, E = DeepgramError> = std::result::Result<T, E>;
 
 impl Deepgram {
     /// Construct a new Deepgram client.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -117,10 +117,6 @@ pub struct Deepgram {
 #[derive(Debug, Error)]
 #[non_exhaustive]
 pub enum DeepgramError {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -154,7 +154,7 @@ pub enum DeepgramError {
 }
 
 #[cfg_attr(not(feature = "listen"), allow(unused))]
-type Result<T> = std::result::Result<T, DeepgramError>;
+type Result<T, E = DeepgramError> = std::result::Result<T, E>;
```
jcdyer, line 157: Making the error type optional means you can `use deepgram::Result;`, but still specify other error types where you need them.

src/listen/websocket.rs
```diff
@@ -37,13 +40,9 @@ use crate::{Deepgram, DeepgramError, Result, Transcription};
 static LIVE_LISTEN_URL_PATH: &str = "v1/listen";
 
 #[derive(Debug)]
-pub struct StreamRequestBuilder<'a, S, E>
-where
-    S: Stream<Item = std::result::Result<Bytes, E>>,
-{
+pub struct StreamRequestBuilder<'a> {
```
jcdyer, line 43: The changes in this file move the generics off of the `StreamRequestBuilder`, and onto `StreamRequest`.  

Previously, `StreamRequestBuilder` carried a `stream: Option<S>`, which was None until the Builder was initialized with a stream.  But the type declaration forced the user to effectively "choose" a stream type when the builder is first created.  In order to support using the API without a stream (manually driving request and response messages in a loop), we actually want a case where there's never a stream.  So now the builder doesn't require that type, and when you choose to add a stream to it, you get a new type (`StreamRequest`) which contains a non-optional `stream: S`.  

In addition to unlocking the non-streaming case, this prevents misuse where a user might try to start the stream before it is initialized.

bd-g: Took me a few reads, but I think I follow. I assume we'll use the same pattern for TTS WS then, when that comes up.

bd-g (approved):

src/listen/websocket.rs
```diff
@@ -267,35 +270,34 @@ where
...
             }
-            if let Some(vad_events) = self.vad_events {
+            if let Some(vad_events) = self.builder.vad_events {
                 pairs.append_pair("vad_events", &vad_events.to_string());
             }
         }
 
         let mut source = self
-            .source
-            .ok_or(DeepgramError::NoSource)?
```
jcdyer, line 298: With the new typing, this error is unnecessary.  You can't call `.start()` until you've attached a source.  (I'll remove that error type in just a moment.

PR #81, opened 2024-07-26
Fix url handling

jcdyer:
## Proposed changes

The current implementation double-escapes query parameters passed in `crate::common::Options`.  This uses proper formencoded parsing instead of string splitting, so the escaping only happens once.

I also introduce a pub fn urlencoded to match the other api endpoints, and to help demonstrate the erroneous behavior.

## Types of changes

## Checklist

### DIFF
--- a/src/listen/websocket.rs
+++ b/src/listen/websocket.rs
@@ -8,40 +8,45 @@
 //!
 //! [api]: https://developers.deepgram.com/api-reference/#transcription-streaming
 
-use crate::common::stream_response::StreamResponse;
-use serde_urlencoded;
-use std::borrow::Cow;
-use std::error::Error;
-use std::fmt::Debug;
-use std::marker::PhantomData;
-use std::path::Path;
-use std::pin::Pin;
-use std::sync::Arc;
-use std::task::{Context, Poll};
-use std::time::Duration;
+use std::{
+    error::Error,
+    fmt::Debug,
+    marker::PhantomData,
+    path::Path,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+    time::Duration,
+};
 
 use bytes::{Bytes, BytesMut};
-use futures::channel::mpsc::{self, Receiver};
-use futures::stream::StreamExt;
-use futures::{SinkExt, Stream};
+use futures::{
+    channel::mpsc::{self, Receiver},
+    stream::StreamExt,
+    SinkExt, Stream,
+};
 use http::Request;
 use pin_project::pin_project;
-use tokio::fs::File;
-use tokio::sync::Mutex;
-use tokio::time;
+use serde_urlencoded;
+use tokio::{fs::File, sync::Mutex, time};
 use tokio_tungstenite::tungstenite::protocol::Message;
 use tokio_util::io::ReaderStream;
 use tungstenite::handshake::client;
 use url::Url;
 
-use crate::common::options::{Encoding, Endpointing, Options};
-use crate::{Deepgram, DeepgramError, Result, Transcription};
+use crate::{
+    common::{
+        options::{Encoding, Endpointing, Options},
+        stream_response::StreamResponse,
+    },
+    Deepgram, DeepgramError, Result, Transcription,
+};
 
 static LIVE_LISTEN_URL_PATH: &str = "v1/listen";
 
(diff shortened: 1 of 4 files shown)


src/listen/websocket.rs
```diff
@@ -69,18 +68,18 @@ impl Transcription<'_> {
...
+    pub fn stream_request_with_options<E, S>(
+        &self,
+        options: Options,
+    ) -> StreamRequestBuilder<'_, S, E>
     where
         S: Stream<Item = std::result::Result<Bytes, E>>,
     {
         StreamRequestBuilder {
-            config: self.0,
+            deepgram: self.0,
```
jcdyer, line 82: I changed this because I was getting confused about what a config is vs an options.

src/listen/websocket.rs
```diff
@@ -233,66 +339,13 @@ impl<'a> StreamRequestBuilder<'a, Receiver<Result<Bytes>>, DeepgramError> {
...
-                            Cow::from(split.next().unwrap_or_default()),
-                            Cow::from(split.next().unwrap_or_default()),
-                        )
-                    })
-                    .collect();
-
-                for (key, value) in query_pairs {
-                    pairs.append_pair(&key, &value);
-                }
-            }
```
jcdyer, line 269: This section got changed above to use actual urlencoded parsing, which prevents a bug in which parameters get double quoted.  If you create an `Options` with a value that includes an `&`, that shows up as `%26` when you serialize `deepgram::common::Options`, but if you pass it to a StreamRequestBuilder`, it serializes as `%2526` (where `%25` is an escaped form of `%` itself).

src/listen/websocket.rs
```diff
@@ -142,10 +144,110 @@ impl Stream for FileChunker {
...
+    ///         DetectLanguage,
+    ///         Encoding,
+    ///         Model,
+    ///         Options,
+    ///     },
+    /// };
+    /// # let mut need_token = std::env::var("DEEPGRAM_API_TOKEN").is_err();
+    /// # if need_token {
+    /// #     std::env::set_var("DEEPGRAM_API_TOKEN", "abc")
+    /// # }
```
jcdyer, line 166: I don't love this pattern.  The `#`s ensure it doesn't show up in the docs, but it's still kind of ugly.

src/listen/websocket.rs
```diff
@@ -248,54 +351,7 @@ where
...
-                            Cow::from(split.next().unwrap_or_default()),
-                            Cow::from(split.next().unwrap_or_default()),
-                        )
-                    })
-                    .collect();
-
-                for (key, value) in query_pairs {
-                    pairs.append_pair(&key, &value);
-                }
-            }
```
jcdyer, line 272: This was the original logic which double-quoted inputs.

src/listen/websocket.rs
```diff
@@ -65,15 +70,12 @@ struct FileChunker {
...
     pub fn stream_request(&self) -> StreamRequestBuilder<'_> {
-        self.stream_request_with_options(None)
+        self.stream_request_with_options(Options::builder().build())
     }
 
-    pub fn stream_request_with_options<'a>(
-        &'a self,
-        options: Option<&'a Options>,
-    ) -> StreamRequestBuilder<'a> {
+    pub fn stream_request_with_options(&self, options: Options) -> StreamRequestBuilder<'_> {
```
jcdyer, line 76: Callout: I changed this public interface to take a `crate::common::options::Options` instead of an `Option<&crate::common::options::Options>`.  It says it takes options, so only sometimes taking them is weird, especially if we also provide a version that doesn't take them, and since we can get the same behavior by passing an empty `Options`.

jcdyer: Just checked:  This function was not available in 0.5.0, so this isn't a breaking change.

src/listen/websocket.rs
bd-g: What is the intent of those lines?

bd-g (approved):

src/listen/websocket.rs
DamienDeepgram: I had a hard time fixing these imports in the examples in VSCode when they would break the build since there is no auto complete for them.

DamienDeepgram (commented):
LGTM

PR #82, opened 2024-08-07
Create a low-level interface for websocket transcription

jcdyer:
- **Add low-level websocket handle to transcription api**
- **fmt**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.6.0] - 2024-07-23
+## [Unreleased](https://github.com/deepgram-devs/deepgram-rust-sdk/compare/0.6.0...HEAD)
+
+## [0.6.0](https://github.com/deepgram-devs/deepgram-rust-sdk/compare/0.5.0...0.6.0) - 2024-08-08
 
 ### Migrating from 0.4.0 -> 0.6.0
 
@@ -124,13 +126,14 @@ Some Enums have changed and may need to be updated
 - custom_topics
 - custom_topic_mode
 
-## [0.5.0]
+## [0.5.0](https://github.com/deepgram-devs/deepgram-rust-sdk/compare/0.4.0...0.5.0) - 2024-07-08
+
 - Deprecate tiers and add explicit support for all currently available models.
 - Expand language enum to include all currently-supported languages.
 - Add (default on) feature flags for live and prerecorded transcription.
 - Support arbitrary query params in transcription options.
 
-## [0.4.0] - 2023-11-01
+## [0.4.0](https://github.com/deepgram-devs/deepgram-rust-sdk/compare/0.3.0...0.4.0) - 2023-11-01
 
 ### Added
 - `detect_language` option.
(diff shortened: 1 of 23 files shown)


CHANGELOG.md
```diff
@@ -8,9 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
```
bd-g, line 8: It won't let me make a suggestion since you didn't edit the line, but let's fix the Markdown hyperlinks at the bottom of the `README`.

```
[Unreleased]: https://github.com/deepgram-devs/deepgram-rust-sdk/compare/0.6.0...HEAD
[0.6.0]: https://github.com/deepgram-devs/deepgram-rust-sdk/compare/0.5.0...0.6.0
[0.5.0]: https://github.com/deepgram-devs/deepgram-rust-sdk/compare/0.4.0...0.5.0
[0.4.0]: https://github.com/deepgram-devs/deepgram-rust-sdk/compare/0.3.0...0.4.0
[0.3.0]: https://github.com/deepgram-devs/deepgram-rust-sdk/compare/0.2.1...0.3.0
```

bd-g (dismissed):
Feels much cleaner.

CHANGELOG.md
jcdyer: I moved these up to the header lines, so they don't get forgotten in the future.

bd-g (changes requested):
The code itself looks clean, something here is breaking the `simple_stream` example - I can get it to run successfully on `main`, but not this branch.

With that said, we also need to update `examples/README.md`. The filename is incorrect, we should suggest exporting an API key, and we should update the example names.

---
## Running Examples

### Setting Env Vars

```sh
export FILENAME=./examples/bueller.wav
# If using the hosted Deepgram API
export DEEPGRAM_API_KEY=<your_api_key_here>
```

### Running the examples

```sh
cargo run --example prerecorded_from_file
cargo run --example callback
cargo run --example make_prerecorded_request_builder
cargo run --example prerecorded_from_url
cargo run --example simple_stream
cargo run --example microphone_stream
cargo run --example text_to_speech_to_file
cargo run --example text_to_speech_to_stream
```

src/speak/rest.rs
```diff
@@ -146,7 +146,7 @@ mod tests {
 
     #[test]
     fn listen_url() {
-        let dg = Deepgram::new("token");
+        let dg = Deepgram::new("token").unwrap();
```
DamienDeepgram, line 149: This is a concern called out in the tracking doc on item #8

jcdyer: Yep.  I updated Deepgram::new() to return a `Result<Deepgram>`, so the caller can decide whether to panic.  Right here, tests `.unwrap()` to signal failure, so this is an appropriate use.

src/listen/websocket.rs
```diff
@@ -366,98 +637,186 @@ where
                 .header("upgrade", "websocket")
                 .header("sec-websocket-version", "13");
 
-            let builder = if let Some(api_key) = self.builder.deepgram.api_key.as_deref() {
-                builder.header("authorization", format!("token {}", api_key))
+            let builder = if let Some(api_key) = builder.deepgram.api_key.as_deref() {
+                http_builder.header("authorization", format!("token {}", api_key))
```
DamienDeepgram, line 641: Should this be capital T for Token?

jcdyer: This way seems to work, but I think it's more usual to capitalize it.  I have updated.

DamienDeepgram:
LGTM

bd-g (approved):
Paired with Cliff to review, committed requested changes.

One notable *TODO* that remains: the stream loop assumes that the `Stream` providing audio to send into the websocket will only send a single `None` message when completed. Certain stream implementations, like `futures::channel`, will provide infinite `None` messages when polled, and this will cause a hot loop in the websocket handler. This will be handled more carefully in a future PR, and for now, SDK users are encouraged to use channels with clean closing behavior, like `tokio::sync::mpsc`.

PR #85, opened 2024-08-12
Streaming callbacks

bd-g:
## Proposed changes

Allows specifying a callback for streaming requests. See [docs](https://developers.deepgram.com/docs/callback#streaming-audio).

## Types of changes

## Checklist

## Further comments

Added a doc to the `callback_method` option to clarify the option is only respected for pre-recorded requests. May be worth making a breaking change in the future and refusing to accept streaming requests that attempt to set this option?

### DIFF
--- a/examples/transcription/websocket/callback_stream.rs
+++ b/examples/transcription/websocket/callback_stream.rs
@@ -45,6 +45,7 @@ async fn main() -> Result<(), DeepgramError> {
         .file(PATH_TO_FILE, AUDIO_CHUNK_SIZE, FRAME_DELAY)
         .await?;
 
+    println!("Deepgram Request ID: {}", results.request_id());
     while let Some(result) = results.next().await {
         println!("got: {:?}", result);
     }
--- a/examples/transcription/websocket/callback_stream.rs
+++ b/examples/transcription/websocket/callback_stream.rs
@@ -1,0 +1,53 @@
+use std::env;
+use std::time::Duration;
+
+use futures::stream::StreamExt;
+
+use deepgram::{
+    common::options::{Encoding, Endpointing, Language, Options},
+    Deepgram, DeepgramError,
+};
+
+static PATH_TO_FILE: &str = "examples/audio/bueller.wav";
+static AUDIO_CHUNK_SIZE: usize = 3174;
+static FRAME_DELAY: Duration = Duration::from_millis(16);
+
+#[tokio::main]
+async fn main() -> Result<(), DeepgramError> {
+    let deepgram_api_key =
+        env::var("DEEPGRAM_API_KEY").expect("DEEPGRAM_API_KEY environmental variable");
+
+    let dg_client = Deepgram::new(&deepgram_api_key)?;
+
+    let options = Options::builder()
+        .smart_format(true)
+        .language(Language::en_US)
+        .build();
+
+    let callback_url = env::var("DEEPGRAM_CALLBACK_URL")
+        .expect("DEEPGRAM_CALLBACK_URL environmental variable")
+        .parse()
+        .expect("DEEPGRAM_CALLBACK_URL not a valid URL");
+
+    let mut results = dg_client
+        .transcription()
+        .stream_request_with_options(options)
+        .keep_alive()
+        .encoding(Encoding::Linear16)
+        .sample_rate(44100)
+        .channels(2)
+        .endpointing(Endpointing::CustomDurationMs(300))
+        .interim_results(true)
+        .utterance_end_ms(1000)
+        .vad_events(true)
+        .no_delay(true)
+        .callback(callback_url)
+        .file(PATH_TO_FILE, AUDIO_CHUNK_SIZE, FRAME_DELAY)
+        .await?;
+
+    while let Some(result) = results.next().await {
+        println!("got: {:?}", result);
+    }
+
+    Ok(())
+}
--- a/src/common/options.rs
+++ b/src/common/options.rs
@@ -1887,7 +1887,12 @@ impl OptionsBuilder {
     ///
     /// See the [Deepgram Callback Method feature docs][docs] for more info.
     ///
+    /// Note that modifying the callback method is only available for pre-recorded audio.
+    /// See the [Deepgram Callback feature docs for streaming][streaming-docs] for details
+    /// on streaming callbacks.
+    ///
     /// [docs]: https://developers.deepgram.com/docs/callback#pre-recorded-audio
+    /// [streaming-docs]: https://developers.deepgram.com/docs/callback#streaming-audio
     ///
     /// # Examples
     ///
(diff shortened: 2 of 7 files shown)


examples/transcription/websocket/callback_stream.rs
```diff
@@ -0,0 +1,53 @@
...
+        .keep_alive()
+        .encoding(Encoding::Linear16)
+        .sample_rate(44100)
+        .channels(2)
+        .endpointing(Endpointing::CustomDurationMs(300))
+        .interim_results(true)
+        .utterance_end_ms(1000)
+        .vad_events(true)
+        .no_delay(true)
+        .callback(callback_url)
```
DamienDeepgram, line 44: I think Cliff wanted to avoid calling it like this due to the different response format which would require to make everything optional.

Response for a callback is a single request ID as far as I know.

Worth checking if this is the case with streaming also

Code for handling callbacks like this for prerecorded which was reverted

https://github.com/deepgram-devs/deepgram-rust-sdk/pull/74/commits/868e2a72b4669aa63ef5caeaab1be4944015ef5e#diff-0b67a8008948b24f9efed7d0976098a96455c5647d48df027f54d739c7972a3eR18-R25

Example of how pre-recorded support the callback url

https://github.com/deepgram-devs/deepgram-rust-sdk/blob/main/examples/transcription/rest/callback.rs#L32

examples/transcription/websocket/callback_stream.rs
```diff
@@ -0,0 +1,53 @@
+use std::env;
```
jcdyer, line 1: This example seems generally harder to set up than other more self-contained ones since it requires the user to have access to a public url that can receive callbacks (or a non-public URL and a self-hosted instance of deepgram).  Maybe add some documentation at the top here suggesting that the caller can provide a presigned POST URL to create an S3 object.

examples/transcription/websocket/callback_stream.rs
jcdyer, line 44: Remind me, do streaming requests still return the same kinds of messages to the original caller when a callback is provided?

src/common/options.rs
```diff
@@ -1887,7 +1887,12 @@ impl OptionsBuilder {
     ///
     /// See the [Deepgram Callback Method feature docs][docs] for more info.
     ///
+    /// Note that modifying the callback method is only available for pre-recorded audio.
+    /// See the [Deepgram Callback feature docs for streaming][streaming-docs] for details
+    /// on streaming callbacks.
```
jcdyer, line 1892: Why is that?  At least when the provided URL is a non-websocket (`http` or `https`) connection, couldn't we use any method the user wants?  Seems like that would be a pretty easy and obvious feature lift in stem.

jcdyer (approved):

examples/transcription/websocket/callback_stream.rs
bd-g: I'm not sure if the presigned POST will work, since we will be sending multiple POST requests back. I just used `ngrok` - would posting a link to their docs be sufficient?

examples/transcription/websocket/callback_stream.rs
bd-g: Some, but not all. I thought that it returned a request ID, but I think it may come back in the headers of the HTTP upgrade. I want to be able to return this, so need to play with this PR more.

The websocket itself actually sends `SpeechSegmentStarted` back to the original caller, everything else to the callback. Which makes sense, in most cases if you have a websocket open, you want to know when the user starts from the calling code so you can handle barge-in and stuff like that.

src/common/options.rs
bd-g: I agree it seems like a "pretty easy and obvious feature lift" in the API. I can only confirm that this is in fact the current behavior, and changing that seems out of scope for this specific PR, so I didn't tackle that yet.

examples/transcription/websocket/callback_stream.rs
jcdyer: Oh does it send a separate POST for each WS message.  That's awkward.  ngrok would only work if you have a self-hosted instance, right?  Maybe we should just leave it alone.

bd-g: Yep, separate POST request each message. And yea, have to setup a local HTTPserver locally, a bit of a pain.

examples/transcription/websocket/callback_stream.rs
jcdyer: Maybe we can drop a request_id into the WebsocketHandle, and provide a `request_id()` method on all the interface types. (`WebsocketHandle` and `RequestStream`).

bd-g: @jcdyer I implemented a way to pull a request ID from a websocket - see the updated examples and the `TranscriptionStream::request_id` method.

We could eventually extend this to batch STT and TTS as well, but going to keep the scope limited to what is needed for the client, and focus on websockets for now.

examples/transcription/websocket/callback_stream.rs
```diff
@@ -0,0 +1,54 @@
...
+        .endpointing(Endpointing::CustomDurationMs(300))
+        .interim_results(true)
+        .utterance_end_ms(1000)
+        .vad_events(true)
+        .no_delay(true)
+        .callback(callback_url)
+        .file(PATH_TO_FILE, AUDIO_CHUNK_SIZE, FRAME_DELAY)
+        .await?;
+
+    println!("Deepgram Request ID: {}", results.request_id());
```
bd-g, line 48: See here - simple way to grab request ID from the `TranscriptionStream`

src/listen/websocket.rs
```diff
@@ -737,6 +759,10 @@ impl<'a> WebsocketHandle {
         // eprintln!("<handle> receiving response: {resp:?}");
         resp
     }
+
+    pub fn request_id(&self) -> Uuid {
```
jcdyer, line 763: Add docs to this function, and please mention that the request ID is needed for support on issues that come up with a particular request.

src/lib.rs
```diff
@@ -172,6 +172,10 @@ pub enum DeepgramError {
     /// An unexpected error occurred in the client
     #[error("an unepected error occurred in the deepgram client: {0}")]
     InternalClientError(anyhow::Error),
+
+    /// A Deepgram API server response was not in the expected format.
+    #[error("The Deepgram API server response was not in the expected format: {0}")]
+    UnexpectedServerResponse(String),
```
jcdyer, line 178: Change this from String to `anyhow::Error`, which is more flexible, and allows us to provide better-structured errors.  (Then at error construction time, you can use `anyhow!("missing request request_id")` and `anyhow!("malformed request_id: {:?}", request_id)`)

jcdyer (changes requested):

src/lib.rs
bd-g: fe7e71d10f9d

src/listen/websocket.rs
bd-g: fe7e71d10f9d

jcdyer (approved):
Looks good!

PR #86, opened 2024-08-15
Implement `From<String>` for `Model`, `Language`, and `Redact`

jcdyer:
## Proposed changes

Add an option to convert strings to various enumerated query options infallibly using `From<String>`.

## Types of changes

## Checklist

## Further comments

A customer has requested this feature.  It is not the default way to specify the options, because we want to encourage customers to do it in a type-safe manner using the enum where possible.  To use this version, one can simply do:

```rust
Options::builder()
    .model(String::from("nova-2").into())
    .build()
```

(Normally, one would probably have a `String` to begin with, if it were coming from a runtime input value, so the `String::from` call would not be needed.  If it is being created directly in code, the existing enum should be used directly.

```rust
Options::builder()
    .model(Model::Nova2)
    .build()
```

### DIFF
--- a/src/common/options.rs
+++ b/src/common/options.rs
@@ -2227,10 +2227,10 @@ impl AsRef<str> for Model {
 impl From<String> for Model {
     fn from(value: String) -> Self {
         match &*value {
-            "nova-2" => Self::Nova2,
-            "nova" => Self::Nova,
-            "enhanced" => Self::Enhanced,
-            "base" => Self::Base,
+            "nova-2" | "nova-2-general" => Self::Nova2,
+            "nova" | "nova-general"  => Self::Nova,
+            "enhanced" | "enhanced-general" => Self::Enhanced,
+            "base" | "base-general" => Self::Base,
             "nova-2-meeting" => Self::Nova2Meeting,
             "nova-2-phonecall" => Self::Nova2Phonecall,
             "nova-2-finance" => Self::Nova2Finance,
--- a/src/common/options.rs
+++ b/src/common/options.rs
@@ -2425,6 +2425,32 @@ fn models_to_string(models: &[Model]) -> String {
 }
 
 #[cfg(test)]
+mod from_string_tests {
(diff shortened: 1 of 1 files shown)


src/common/options.rs
```diff
@@ -2224,6 +2224,52 @@ impl AsRef<str> for Model {
     }
 }
 
+impl From<String> for Model {
+    fn from(value: String) -> Self {
+        match &*value {
+            "nova-2" => Self::Nova2,
+            "nova" => Self::Nova,
+            "enhanced" => Self::Enhanced,
+            "base" => Self::Base,
```
bd-g, line 2233: Let's explicitly handle `*-general` model specifiers. 
```suggestion
            "nova-2" | "nova-2-general" => Self::Nova2,
            "nova" | "nova-general"  => Self::Nova,
            "enhanced" | "enhanced-general" => Self::Enhanced,
            "base" | "base-general" => Self::Base,
```

bd-g (approved):

src/common/options.rs
```diff
@@ -2286,15 +2332,86 @@ impl AsRef<str> for Language {
...
 impl AsRef<str> for Redact {
     fn as_ref(&self) -> &str {
-        use Redact::*;
-
         match self {
-            Pci => "pci",
-            Numbers => "numbers",
-            Ssn => "ssn",
-            Other(id) => id,
+            Redact::Pci => "pci",
```
DamienDeepgram, line 2400: Redaction supports 50 lower level Entity Types also

https://developers.deepgram.com/docs/supported-entity-types

jcdyer: Cool.  Let's create an issue to support those explicitly.  That's out of scope for this PR.

jcdyer: (The issue: https://github.com/deepgram-devs/deepgram-rust-sdk/issues/87)

DamienDeepgram:
LGTM

PR #103, opened 2025-01-13
Set host header based on host URL instead of hardcoding it

RyanChimienti:
Attempting to make a streaming listen request to my hosted server was failing because the SDK was setting the `host` header as `api.deepgram.com`, but the Deepgram instance I was running lived at a different domain. This PR sets the `host` header to whichever host is actually receiving the request.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "deepgram"
-version = "0.6.4"
+version = "0.6.5"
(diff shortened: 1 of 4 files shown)


RyanChimienti:
Looks like clippy and fmt updates hit recently which was making CI fail. I ran those as well.

bd-g (approved):

PR #106, opened 2025-01-24
Fix select biased by fusing streams outside loop

jcdyer:
- Fuse streams properly in select_biased in `WebsocketHandle::stream` and `run_worker`.

Based on [the documentation for select_biased](https://docs.rs/futures-util/latest/futures_util/macro.select_biased.html#:~:text=call%20is%20in%20a%20loop,every%20branch%20in%20a%20select_biased!), as reported by a customer, streams used in select_biased need to be fused on the stream, not the future, and futures polled more than once need to be fused outside the loop.  Otherwise, the inner future will need to be re-polled each time to determine whether fusing needs to happen, which causes a busy loop (at best) in the case of Streams, and a possible panic for Futures.

### DIFF
--- a/src/listen/websocket.rs
+++ b/src/listen/websocket.rs
@@ -462,15 +462,17 @@ macro_rules! send_message {
 async fn run_worker(
     ws_stream: WebSocketStream<MaybeTlsStream<tokio::net::TcpStream>>,
     mut message_tx: Sender<WsMessage>,
-    mut message_rx: Receiver<WsMessage>,
+    message_rx: Receiver<WsMessage>,
     mut response_tx: Sender<Result<StreamResponse>>,
     keep_alive: bool,
 ) -> Result<()> {
     // We use Vec<u8> for partial frames because we don't know if a fragment of a string is valid utf-8.
     let mut partial_frame: Vec<u8> = Vec::new();
-    let (mut ws_stream_send, mut ws_stream_recv) = ws_stream.split();
+    let (mut ws_stream_send, ws_stream_recv) = ws_stream.split();
+    let mut ws_stream_recv = ws_stream_recv.fuse();
     let mut is_open: bool = true;
     let mut last_sent_message = tokio::time::Instant::now();
+    let mut message_rx = message_rx.fuse();
     loop {
         // eprintln!("<worker> loop");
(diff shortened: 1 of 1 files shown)


src/listen/websocket.rs
```diff
@@ -461,15 +462,17 @@ macro_rules! send_message {
...
     keep_alive: bool,
 ) -> Result<()> {
     // We use Vec<u8> for partial frames because we don't know if a fragment of a string is valid utf-8.
     let mut partial_frame: Vec<u8> = Vec::new();
-    let (mut ws_stream_send, mut ws_stream_recv) = ws_stream.split();
+    let (mut ws_stream_send, ws_stream_recv) = ws_stream.split();
+    let mut ws_stream_recv = ws_stream_recv.fuse();
     let mut is_open: bool = true;
     let mut last_sent_message = tokio::time::Instant::now();
+    let mut message_rx = message_rx.fuse();
```
bd-g, line 475: It won't let me comment on the `tokio::time::sleep_until` variable a few lines down.

I presume this doesn't need to be fused outside the loop, since the variable is created within the loop, so it isn't subject to the same polling issues as the other futures, which are called repeatedly within the loop?

bd-g: Also, since `message_rx` is fused here, shouldn't we *not* fuse it within the loop?

It [seems like](https://github.com/rust-lang/rust-clippy/issues/11143) `object.next().fuse()` can override the `object.fuse()` call and still cause rough behavior.

bd-g (changes requested):

src/listen/websocket.rs
jcdyer: Correct about sleep.  Each new iteration of the loop gets a new future, so any given sleep future will not get polled after the single `select_biased!` construct completes.

jcdyer: Good catch on the `message_rx.next().fuse()`.  Fixed in 35b2083

jfoster-twilio (commented):
This is pretty close to what we were testing -- just a couple suggestion on unnecessary .fuse calls

src/listen/websocket.rs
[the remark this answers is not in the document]
jcdyer: Oh, good catch.  Removed in 35b2083

bd-g (approved):

PR #114, opened 2025-06-25
move dependencies to dev

jcaden:
- **Move audio and rodio dependencies to dev-dependencies**
- **Remove unused dependencies**


jcdyer:
Feel free to ignore or suppress the clippy errors, as they are new with the latest rust release.  We will resolve them soon, but do note that we have some transitive dependencies specified to satisfy minimal-versions checks, and those should not be removed.

jcaden:
I modified the patches. I tested locally all the test that failed and seems to work now.

PR #115, opened 2025-06-25
Solve result too large issue

jcaden:
Error code is too big because TunsgteniteError is too large this patch proposes using a Box to contain reducing the type error using the heap instead of the stack. I've tested against our code that makes use of this error and it does not seems to affect compilation.

---

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,7 +148,7 @@ pub enum DeepgramError {
     #[cfg(feature = "listen")]
     /// Something went wrong with WS.
     #[error("Something went wrong with WS: {0}")]
-    WsError(#[from] TungsteniteError),
+    WsError(#[from] Box<TungsteniteError>),
 
     /// Something went wrong during serialization/deserialization.
     #[error("Something went wrong during json serialization/deserialization: {0}")]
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -148,7 +148,7 @@ pub enum DeepgramError {
     #[cfg(feature = "listen")]
     /// Something went wrong with WS.
     #[error("Something went wrong with WS: {0}")]
-    WsError(#[from] TungsteniteError),
+    WsError(#[from] Box<TungsteniteError>),
```
jcdyer, line 151: This is a semver breaking change for anyone trying to pattern match into this error type.  Not a big deal, but we may want to delay merging this PR until we want to make a breaking change for other reasons.  In the meantime, I'd be okay with suppressing the lint.

thisislvca:
Thx for the changes Jose, very appreciated as an user of this crate 🙏

jcdyer (approved):
This is a good fix, but we'll put it behind a minor (pre-1.0 semver-incompatible) version bump, because it breaks code that tries to match through the tungstenite error.  Here's an example that breaks on this branch, but not on the currently released (0.6.8) SDK.

```rust
use deepgram::DeepgramError;
use tungstenite::Error as TungsteniteError;
fn match_error(err: DeepgramError) {
    if let DeepgramError::WsError(TungsteniteError::AlreadyClosed) = err {
        println!("socket already closed");
    }
}
```

Since rust provides no way to directly match through a box, this code would need to become something like:

```rust
use deepgram::DeepgramError;
use tungstenite::Error as TungsteniteError;
fn match_error(err: DeepgramError) {
    if let DeepgramError::WsError(tungerr)  = err {
        if let TungsteniteError::AlreadyClosed = &*tungerr {
            println!("socket already closed");
        }
    }
}
```

PR #116, opened 2025-06-25
feat: add support for token-based authentication

thisislvca:
## Proposed changes

This PR adds proper support for temporary token authentication to the Deepgram Rust SDK. I'm working on a client-side application, and I was surprised to discover that the SDK only supported API key authentication with the "Token" prefix, thus not supporting [token-based auth](https://developers.deepgram.com/guides/fundamentals/token-based-authentication).

- **API Keys**: Should use the `Token` prefix (e.g., `Token dg_xxx`)
- **Temporary Tokens**: Should use the `Bearer` prefix (e.g., `Bearer dg_xxx`)

The existing implementation incorrectly used "Token" for all authentication, which worked for API keys but was wrong for temporary tokens.

**Changes:**

There are no breaking changes to current apps.

- Added `AuthMethod` enum to distinguish between API keys and temporary tokens
- Updated `Deepgram` struct to use the new authentication system (by default it'll still use API keys)
- Added `Deepgram::with_temp_token()` constructor for temporary token authentication
- Use new auth logic in websocket
- Added authentication documentation to README
- Added tests for the new authentication logic

All existing tests pass, and the change is backwards compatible for API usage patterns.

## Types of changes

- [x] Documentation update or tests (if none of the other choices apply)

## Checklist

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 /Cargo.lock
 your_output_file.wav
+.DS_Store
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -105,13 +105,35 @@ impl Deref for RedactedString {
     }
 }
 
+/// Authentication method for Deepgram API requests.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AuthMethod {
+    /// Use an API key with "Token" prefix (e.g., "Token dg_xxx").
+    /// This is for permanent API keys created in the Deepgram console.
+    ApiKey(RedactedString),
+
+    /// Use a temporary token with "Bearer" prefix (e.g., "Bearer dg_xxx").
+    /// This is for temporary tokens obtained via token-based authentication.
+    TempToken(RedactedString),
+}
+
+impl AuthMethod {
+    /// Get the authorization header value for this authentication method.
+    pub(crate) fn header_value(&self) -> Option<String> {
+        match self {
+            AuthMethod::ApiKey(key) => Some(format!("Token {}", key.0)),
+            AuthMethod::TempToken(token) => Some(format!("Bearer {}", token.0)),
+        }
+    }
+}
+
 /// A client for the Deepgram API.
 ///
 /// Make transcriptions requests using [`Deepgram::transcription`].
 #[derive(Debug, Clone)]
 pub struct Deepgram {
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -89,7 +89,8 @@ impl Transcription<'_> {
 }
 
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
-pub(crate) struct RedactedString(pub String);
+/// A string wrapper that redacts its contents when formatted with `Debug`.
+pub struct RedactedString(pub String);
```
jcdyer, line 93: This should remain `pub(crate)`

src/lib.rs
```diff
@@ -105,13 +106,35 @@ impl Deref for RedactedString {
     }
 }
 
+/// Authentication method for Deepgram API requests.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AuthMethod {
```
jcdyer, line 111: This should also be `pub(crate)`.  It does not need to be exposed to the end user.

src/lib.rs
```diff
@@ -105,13 +106,35 @@ impl Deref for RedactedString {
...
+}
+
+impl AuthMethod {
+    /// Get the authorization header value for this authentication method.
+    pub(crate) fn header_value(&self) -> Option<String> {
+        match self {
+            AuthMethod::ApiKey(key) => Some(format!("Token {}", key.0)),
+            AuthMethod::TempToken(token) => Some(format!("Bearer {}", token.0)),
+        }
+    }
```
jcdyer, line 128: This always returns `Some()` so it can return a `String instead of an `Option<String>`.

jcdyer (changes requested):
Thanks for this contribution.  It looks good.  I don't think we want to commit to supporting `RedactedString` or `AuthMethod` as part of the public API at this point.

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 /Cargo.lock
-your_output_file.wav
\ No newline at end of file
+your_output_file.wav
+.DS_Store
```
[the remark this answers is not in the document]
thisislvca: Yeah it's mostly for people who don't have it. Most gitignore templates have it too :)

jcdyer (approved):

PR #117, opened 2025-06-29
Update dependencies to latest version

jcaden:
---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,36 +20,37 @@ categories = ["api-bindings", "multimedia::audio"]
 [dependencies]
 # TODO Investigate which of these dependencies can go behind features.
 bytes = "1"
-futures = "0.3"
-http = "0.2"
+futures = "^0.3"
+http = "1.3"
 pin-project = "1"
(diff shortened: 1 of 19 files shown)


jcdyer (changes requested):
I jumped the gun on my review. We need to make sure the Minimal Versions still pass.  Usually this means bumping patch versions on the version specifier for the crate in question (or the crate that introduced a transitive dependency that isn't building).  Occasionally, it means adding a version override the the transitive dependency directly.  See the section beginning on line 44 of Cargo.toml

jcaden:
Thanks for the comments. I've amend the commit I run the minimal version test locally and it works. I hope it passes on CI system as well..

jpvajda:
It looks like the Clippy check is failing on this PR after the CI checks were run.

jcdyer:
#115 is merged now.  Can you rebase this PR, and fix the fmt issues if they're still showing up?

jcaden:
Done, thanks for merging #115. Also fixed format.

jcdyer:
CI is now running again here: https://github.com/deepgram/deepgram-rust-sdk/actions/runs/16197851369

PR #125, opened 2025-07-25
Support detected languages for multilingual transcripts

antonshwarts:
When switching between multiple languages, TranscriptResponse should include:
- a list of detected languages for the entire transcript
- the detected language for each individual word

[doc]https://developers.deepgram.com/docs/multilingual-code-switching

## Types of changes

## Checklist

### DIFF
--- a/src/common/stream_response.rs
+++ b/src/common/stream_response.rs
@@ -26,6 +26,9 @@ pub struct Word {
 
     #[allow(missing_docs)]
     pub punctuated_word: Option<String>,
+
+    #[allow(missing_docs)]
+    pub language: Option<String>,
 }
 
 /// Transcript alternatives.
@@ -43,6 +46,9 @@ pub struct Alternatives {
 
     #[allow(missing_docs)]
     pub confidence: f64,
+
+    #[allow(missing_docs)]
+    pub languages: Option<Vec<String>>,
 }
 
 /// Transcription results for a single audio channel.
--- a/src/common/stream_response.rs
+++ b/src/common/stream_response.rs
@@ -48,6 +48,7 @@ pub struct Alternatives {
     pub confidence: f64,
 
     #[allow(missing_docs)]
+    #[serde(default)]
     pub languages: Vec<String>,
 }
 
--- a/src/common/stream_response.rs
+++ b/src/common/stream_response.rs
@@ -48,7 +48,7 @@ pub struct Alternatives {
     pub confidence: f64,
 
     #[allow(missing_docs)]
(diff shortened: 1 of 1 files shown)


src/common/stream_response.rs
```diff
@@ -43,6 +46,9 @@ pub struct Alternatives {
 
     #[allow(missing_docs)]
     pub confidence: f64,
+
+    #[allow(missing_docs)]
+    pub languages: Option<Vec<String>>,
```
bd-g, line 51:
```suggestion

    #[allow(missing_docs)]
    #[serde(default)]
    pub languages: Vec<String>,
```

Let's avoid nesting a vector inside an `Option`. An empty vector communicates the same information as `Option::is_none`.

bd-g (changes requested):

src/common/stream_response.rs
bd-g: The codebase needs to be updated either way if SDK consumers are creating these objects themselves, and not just deserializing from us. Either way, we need to implement `Default` on the struct.

src/common/stream_response.rs
```diff
@@ -26,6 +26,9 @@ pub struct Word {
 
     #[allow(missing_docs)]
     pub punctuated_word: Option<String>,
+
+    #[allow(missing_docs)]
+    pub language: Option<String>,
```
bd-g, line 31: @cpierce-deepgram You need a #[serde(default)] here too.

bd-g (changes requested):

cpierce-deepgram:
Tested the proposed change using `dg_client.stream_request_with_options(options)`:
- With option `language(Language::multi)`: verified languages=["en"] and language="en"
- With option `language(Language::en)`: verified languages=[] and language=None

src/common/stream_response.rs
cpierce-deepgram: It's not necessary to add the default value there. When using `Option<String>` the default value is automatically `None`

cpierce-deepgram:
Tested the proposed change using dg_client.transcription().stream_request_with_options(options):

With option language(Language::multi): verified languages=["en"] and language="en"
With option language(Language::en): verified languages=[] and language=None

src/common/stream_response.rs
bd-g: Right, my fly-by reading was too quick!

bd-g (approved):

jjmaldonis:
Thanks for your PR!

PR #162, opened 2026-07-08
feat: Use feature flag to allow rustls-no-provider with reqwest

SanchayanMaity:
This allows applications or other users of the crate to not bring aws-lc-rs as a dependency due to cargo feature unification and let a user of the crate install a crypto provider manually for use by reqwest.

See https://docs.rs/rustls/latest/rustls/#cryptography-providers for details.

When opting into reqwest/rustls-no-provider, application must install a crypto provider. For example, rustls::crypto::ring::default_provider().install_default().

## Types of changes

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ http = "1.4"
 pin-project = "1"
 reqwest = { version = "^0.13", default-features = false, features = [
   "json",
-  "rustls",
   "stream",
   "query",
 ] }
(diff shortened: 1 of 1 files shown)


SanchayanMaity:
For some more context on this change,

We use the deepgram SDK in [GStreamer](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/) for transcribing. We have avoided pulling in `aws-lc-rs` as a dependency for various reasons.  Because of feature unification in cargo, a dependency enabling `aws-lc-rs` (`reqwest` feature flag `rustls` in this case here) results in it becoming part of the dependency chain.

`reqwest` already allows control over this and so do other crates like `quinn`. The current change should not affect any downstream users while allowing other users leeway in opting for a different crypto provider.

Issue #130, opened 2025-10-31
Extra metadata support


yujonglee:
## Proposed changes

https://github.com/deepgram/deepgram-rust-sdk/blob/d2f2723e64375d771086859d56c2b4434757a2da/src/common/stream_response.rs#L87

Include `extra: Option<HASHMAP>`

## Context

https://developers.deepgram.com/docs/extra-metadata

## Possible Implementation

as Option field

dg-coreylweathers:
Partial status (`52711b7`, v0.10.0):

**Request side — done.** `OptionsBuilder::extra(HashMap<String,String>)` is implemented (`src/common/options.rs:1987`), backed by an `extra: Option<HashMap<String,String>>` field and serialized into the request query params.

**Remaining sliver — streaming response `extra`.** The `extra` field is not yet surfaced on the streaming response type (`src/common/stream_response.rs`). This is queued as a Phase 1 quick-win and I'll close this issue when that lands. Keeping open until then.

dg-coreylweathers:
The remaining streaming-response `extra` sliver is implemented in #164: `extra` is now surfaced on `stream_response::Metadata` and the terminal Metadata message. Will close when v0.11.0 lands on `main`.

Issue #129, opened 2025-09-11
Public paragraph fields of transcription


huykingsofm:
## Proposed changes

Public paragraph fields of transcription (transcript, paragraphs)

## Context

Why I can use the feature if fields are private.

dg-coreylweathers:
Addressed in #164 (Phase 1 quick wins, branch `feat/phase-1-quick-wins`). Will close when the parity release (v0.11.0) lands on `main`.

Issue #128, opened 2025-09-11
Support whisper models


huykingsofm:
## Proposed changes

Support whisper models in speech to text.

## Context

I am running speech to text and using whisper tiny for low cost but not found whisper models in deepgram::common::options::Model.

## Possible Implementation

Add whisper models.

dg-coreylweathers:
Addressed in #164 (Phase 1 quick wins, branch `feat/phase-1-quick-wins`). Will close when the parity release (v0.11.0) lands on `main`.

Issue #87, opened 2024-08-16
Support all redaction options


jcdyer:
## Proposed changes

Add enumerated support for all the redaction options listed at  https://developers.deepgram.com/docs/supported-entity-types

## Context

These are already accessible by providing a `Redact::Other(value)`, but we should support them with explicit enum variants.

## Possible Implementation

N/A

dg-coreylweathers:
Addressed in #164 (Phase 1 quick wins, branch `feat/phase-1-quick-wins`). Will close when the parity release (v0.11.0) lands on `main`.

PR #164, opened 2026-07-31
feat: Phase 1 quick wins — TTS request-id, Whisper models, redaction, response fields, captions

### LINKED ISSUE #89: Add method to get request ID from TTS requests
## Proposed changes

The Deepgram request ID for TTS requests is passed via a header value, not in the output data (which is the audio waveform for the speech). We should expose a top level function to be able to access the request ID, with the SDK handling pulling this from the headers.

## Context

Recorded request IDs is important for debugging and general tracking purposes.

dg-coreylweathers:
## Phase 1 — Quick-win release (Tier C)

Part of the JS-SDK parity program. Bundles the small, self-contained ergonomics/completeness items into one reviewable PR. Each concern is its own commit.

### What's included
- **#89 — TTS request-id / response metadata.** TTS REST discarded the response headers, so `dg-request-id` was unrecoverable. Adds `SpeakMetadata` and `speak_to_file_with_metadata` / `speak_to_stream_with_metadata`; existing methods keep their signatures. Also drops a stray `println!` from `speak_to_file`.
- **#128 — Named Whisper model variants.** `Model::Whisper`, `WhisperTiny/Base/Small/Medium/Large` (were only reachable via `Model::CustomId`).
- **#87 — Extended redaction (partial).** `Redact::Pii`, `Redact::Phi`, `Redact::AggressiveNumbers`; the individual entity-type variants are not included.
- **#129 — Public response fields.** Paragraph & Audio Intelligence response types had private fields with no accessors; now `pub` + `#[non_exhaustive]`.
- **#130 — `extra` on streaming response.** Surfaced on `stream_response::Metadata` and the terminal Metadata message.
- **Captions helper.** `common::captions::srt` / `webvtt` (+ `Response::to_srt`/`to_webvtt`) mirroring `@deepgram/captions`. New `captions` example.

### Verification (run in a Rust 1.97 container)
- `cargo build --all-features --all-targets` ✅
- `cargo test --all --all-features` ✅ (137 passed)
- `cargo clippy --all-features --all-targets -- -D warnings` ✅
- `cargo fmt --all --check` ✅
- `cargo doc --all-features` (`-D warnings`) ✅
- Per-feature `cargo check` (no-default / listen / speak / manage) ✅

### SemVer note
`cargo semver-checks` reports **2 breaking changes**, both from **#130** on deserialize-only streaming types (`Metadata` marked `#[non_exhaustive]`; `extra` added to `Metadata` and `TerminalResponse`). Everything else is additive (verified). These are flagged **BREAKING** in the changelog and are covered by the single accumulated **0.11.0** release at the end of the program (a 0.x minor bump permits breaking changes). Per the program's changelog discipline, `Cargo.toml` is **not** bumped per-PR.

Closes #89, #128, #129, #130. Partially addresses #87 (adds `pii`, `phi`, `aggressive_numbers`; the per-entity-type variants from the Supported Entity Types page remain open).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Text-to-Speech REST responses now expose their metadata headers, including the `dg-request-id`. New `Speak::speak_to_file_with_metadata` and `Speak::speak_to_stream_with_metadata` return a `SpeakMetadata` (request id, model name/uuid, character count, content type) alongside the audio ([#89](https://github.com/deepgram/deepgram-rust-sdk/issues/89)). New example `text_to_speech_request_id`.
+- Named Deepgram Whisper Cloud model variants: `Model::Whisper`, `Model::WhisperTiny`, `Model::WhisperBase`, `Model::WhisperSmall`, `Model::WhisperMedium`, and `Model::WhisperLarge` (previously only reachable via `Model::CustomId`) ([#128](https://github.com/deepgram/deepgram-rust-sdk/issues/128)).
+- Additional redaction options: `Redact::Pii`, `Redact::Phi`, and `Redact::AggressiveNumbers` ([#87](https://github.com/deepgram/deepgram-rust-sdk/issues/87)).
+- Public fields on the Audio Intelligence and paragraph response types (`Paragraph`, `Sentence`, `Paragraphs`, `Entity`, `Intent`, `Segment`, `Intents`, `SentimentSegment`, `SentimentAverage`, `Sentiments`, `TopicDetail`, `TopicSegment`, `Topics`, `Summary`), which were previously inaccessible. All are now `#[non_exhaustive]` ([#129](https://github.com/deepgram/deepgram-rust-sdk/issues/129)).
+- Caption generation: `common::captions::srt` / `webvtt` (and `Response::to_srt` / `to_webvtt`) render SRT and WebVTT subtitle files from a pre-recorded transcription, with configurable words-per-cue via `CaptionOptions`. New example `captions`.
+- `extra` metadata is now surfaced on the streaming response: `common::stream_response::Metadata` and the `StreamResponse::TerminalResponse` (Metadata) message carry the `extra` key-value pairs echoed back from the request ([#130](https://github.com/deepgram/deepgram-rust-sdk/issues/130)).
+
+### Changed
+
+- **BREAKING**: `common::stream_response::Metadata` is now `#[non_exhaustive]` and gained an `extra` field, and `StreamResponse::TerminalResponse` gained an `extra` field (the variant is now `#[non_exhaustive]`). Callers constructing or exhaustively destructuring either with a struct literal must add `..`. (These types are produced by deserialization, so most callers are unaffected.)
+- `Speak::speak_to_file` no longer prints `Audio saved to …` to stdout on success; libraries should not write to stdout. Error diagnostics are unchanged.
+
 ## [0.10.0](https://github.com/deepgram/deepgram-rust-sdk/compare/0.9.2...0.10.0)
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,6 +85,11 @@ path = "examples/transcription/rest/prerecorded_from_url.rs"
 required-features = ["listen"]
 
 [[example]]
+name = "captions"
+path = "examples/transcription/rest/captions.rs"
+required-features = ["listen"]
+
+[[example]]
 name = "simple_stream"
 path = "examples/transcription/websocket/simple_stream.rs"
 required-features = ["listen"]
@@ -137,4 +142,9 @@ required-features = ["speak"]
 [[example]]
 name = "text_to_speech_to_stream"
 path = "examples/speak/rest/text_to_speech_to_stream.rs"
+required-features = ["speak"]
+
+[[example]]
+name = "text_to_speech_request_id"
+path = "examples/speak/rest/text_to_speech_request_id.rs"
 required-features = ["speak"]
--- a/examples/speak/rest/text_to_speech_request_id.rs
+++ b/examples/speak/rest/text_to_speech_request_id.rs
@@ -1,0 +1,48 @@
+//! Generate speech and read the response metadata — including the
+//! `dg-request-id` — using [`Speak::speak_to_file_with_metadata`].
+//!
+//! Run with:
+//!
+//! ```sh
+//! DEEPGRAM_API_KEY=your-key cargo run --example text_to_speech_request_id --features speak
+//! ```
+
+use std::{env, path::Path};
+
+use deepgram::{
+    speak::options::{Container, Encoding, Model, Options},
+    Deepgram, DeepgramError,
+};
+
+#[tokio::main]
+async fn main() -> Result<(), DeepgramError> {
+    let deepgram_api_key =
+        env::var("DEEPGRAM_API_KEY").expect("DEEPGRAM_API_KEY environmental variable");
+
+    let dg_client = Deepgram::new(&deepgram_api_key)?;
+
+    let options = Options::builder()
+        .model(Model::AuraAsteriaEn)
+        .encoding(Encoding::Linear16)
+        .sample_rate(16000)
+        .container(Container::Wav)
+        .build();
+
+    let text = "Hello, how can I help you today?";
+    let output_file = Path::new("your_output_file.wav");
+
+    let metadata = dg_client
+        .text_to_speech()
+        .speak_to_file_with_metadata(text, &options, output_file)
+        .await?;
+
+    // The request id is essential for debugging and support requests.
+    match metadata.request_id() {
+        Some(request_id) => println!("dg-request-id: {request_id}"),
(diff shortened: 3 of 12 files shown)


GregHolmes (changes requested):
# Review: deepgram-rust-sdk #164, feat: Phase 1 quick wins - TTS request-id, Whisper models, redaction, response fields, captions
Verdict: request-changes

## Intent
Add the Tier C Rust SDK parity features: TTS response headers, named Whisper models, additional redaction variants, public batch-response fields, streaming `extra` metadata, and SRT/WebVTT helpers. Existing TTS method signatures must stay stable. The two confirmed public-API breaks are intended for the accumulated 0.11.0 pre-1.0 release. Intent is based on the PR and issues #89, #87, #128, #129, and #130; no launch or Slack context is linked.

## Blocking
- [B1] `src/common/stream_response.rs:165`, terminal `extra` cannot be read by SDK users.

**Title:** [B1] Terminal response hides its new `extra` metadata

**Summary:** The PR and changelog say terminal streaming metadata is surfaced, but `StreamResponse::TerminalResponse::extra` is private. Rust privacy prevents consumers from binding or reading it when they match the terminal message, so this half of issue #130 remains unavailable.

**Expected:** A consumer receiving `StreamResponse::TerminalResponse` can read the echoed `extra` key-value pairs.

**Observed:** The response deserializes the field, but only code inside this crate can access it.

**Recommended fix:** Change the field to `pub extra: Option<HashMap<String, String>>` and add a deserialization test that matches a terminal message containing `extra` from outside the defining module.

- [B2] `src/common/captions.rs:100`, caption grouping loses speaker and utterance boundaries.

**Title:** [B2] Captions do not mirror the stated Deepgram caption helper behavior

**Summary:** The referenced `deepgram-js-captions` helper preserves each returned utterance as a cue and, when utterances are absent, starts a new cue when the diarized speaker changes. This implementation always chunks `channels[0].alternatives[0].words`; a cue can span an utterance pause or include words from another speaker while carrying the first speaker's label.

**Expected:** Caption cues preserve utterance boundaries and do not attribute words to a preceding diarized speaker.

**Observed:** A caption for a speaker change can render as `[Speaker 0] ...words from speaker 1...`, and an utterance boundary can become one long cue.

**Recommended fix:** Match the JS helper's source selection: use `results.utterances` when present, and otherwise flush the buffered cue on a speaker change as well as at `max_words_per_cue`; add focused tests for both cases.

- [B3] `Cargo.toml:3` and `CHANGELOG.md:8`, the branch conflicts with current `main`.

**Title:** [B3] Pull request cannot merge against the current base

**Summary:** GitHub reports `mergeable: CONFLICTING`, and `git merge-tree --write-tree origin/main HEAD` confirms content conflicts in both files. The branch is four commits behind `origin/main`, including the 0.10.1 release, Flux TTS, ForceEndTurn/timings, and connect diagnostics.

**Expected:** The PR rebases cleanly onto the current base and retains both the base's release history and this PR's additions.

**Observed:** GitHub cannot create a merge commit, and the unchanged branch would overwrite current `Cargo.toml` and changelog content during manual resolution.

**Recommended fix:** Rebase `feat/phase-1-quick-wins` onto `origin/main`, resolve both files by preserving the 0.10.1 release and current feature declarations, then rerun the complete CI suite.

## Should-fix
- [S1] `src/speak/response.rs:53`, the new response contracts lack direct regression tests.

**Title:** [S1] New public mappings are not covered by focused tests

**Summary:** The test suite passes, but it has no test for `SpeakMetadata::from_headers`, no terminal-response `extra` deserialization test, and its existing model/redaction tests do not exercise the new Whisper or redaction variants. These mappings are exactly where a header spelling or serialized value can silently drift from the API.

**Expected:** Each new public response field and option serializes or deserializes through a test that asserts its exact wire value.

**Observed:** Existing passing tests cover only pre-existing model/redaction values and caption happy paths.

**Recommended fix:** Add table-driven header tests for all `SpeakMetadata` fields, a terminal `extra` fixture, and exact URL serialization/round-trip assertions for every added `Model` and `Redact` variant.

## Nits
- None.

## Verified (evidence)
- TTS response header names and meanings match the live Text-to-Speech documentation: `dg-request-id`, `dg-model-name`, `dg-model-uuid`, `dg-char-count`, `content-type`, `transfer-encoding`, and `date`.
- Whisper wire values and availability match the live Whisper Cloud documentation: `whisper`, `whisper-tiny`, `whisper-base`, `whisper-small`, `whisper-medium`, and `whisper-large`; the model is pre-recorded only.
- Redaction wire values and descriptions for `pii`, `phi`, and `aggressive_numbers` match the live Supported Entity Types documentation.
- `extra` is documented as a response `metadata.extra` object, consistent with the public `Metadata::extra` field.
- Local Rust 1.94.1: `cargo build --all-features --all-targets`, `cargo test --all --all-features`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo fmt --all --check`, `RUSTDOCFLAGS="-D warnings" cargo doc --all-features`, and all four no-default feature checks pass. The declared `rust:1.97` image was pulled but contains neither `cargo` nor `rustup`, so it could not run the gate.

## Needs human
- Greg: confirm the intended SemVer gate policy for this accumulation. CI correctly finds two breaking changes (`Metadata` and `TerminalResponse`) but compares 0.10.0 to unchanged 0.10.0 and fails; the PR proposes carrying them in a later 0.11.0 release. The changelog is honest and a 0.x minor release can carry the break, but whether this PR may remain red until the aggregation/release commit is a release-process decision.

## Developer-facing messaging
- The changelog accurately calls out the two deliberate `#[non_exhaustive]` breaks, but its statement that terminal responses "carry the `extra` key-value pairs" is false until [B1] makes the field public.
- Once [B2] is fixed, the caption helper can truthfully claim parity with the JavaScript implementation without producing captions that misattribute speakers or bridge utterance pauses.

dg-coreylweathers:
## Round 2 — pushed, with four more fixes

Rebased onto current `main` (0.11.0) and re-reviewed end to end, including live calls against production.

- **The TTS metadata was incomplete.** A live `/v1/speak` request returns four headers the SDK was discarding: `dg-additional-model-uuids`, `dg-project-id`, `dg-breaks-applied`, `dg-pronunciations-applied`. The docs enumerate no TTS response headers at all, which is why a doc-derived field list missed them — the list now comes from an observed response. `additional_model_uuids` arrives comma-separated, so it is split into a `Vec<String>`. This is the same data loss you blocked #166 for on the streaming side.
- **The `**BREAKING**` bullet gave advice that does not compile.** It said callers must "add `..`". `#[non_exhaustive]` rejects a struct literal outright, and functional-record-update too, so a consumer following it hits a second error. It now says to deserialize instead, and keeps `..` for pattern matching, where it does apply.
- **Version is now 0.12.0.** `main` shipped 0.11.0 and this branch carries two signature-level breaks, so the SemVer job was red. Verified the bump clears it: `Checking 0.11.0 -> 0.12.0 (major change) … no semver update required`. `Cargo.lock` and the AGENTS.md version sentence updated with it.
- **`speak_to_file`'s dropped stdout line is marked `**BREAKING**`** — the bullet already told consumers to print it themselves, which is the definition of the break.
- Skill manifests pinned to `0.12`.

**What I checked against the live API**

- **Are the Whisper values real?** Yes — all six return HTTP 200 on `/v1/listen`.
- **Are the redaction values real?** Yes — `pii`, `phi`, `aggressive_numbers` all accepted.
- **Is `extra` actually echoed on the streaming path?** Yes, and this one mattered: it comes back on all 20 interim `Results` messages *and* the terminal `Metadata`. That was worth confirming because if it were absent, the field would be a false promise and one of the two semver breaks would be unnecessary. It is real.
- **B1 re-check:** `TerminalResponse::extra` was never actually private — enum variant fields carry the enum's visibility. The fix is the external-crate test now proving it, which is the right outcome either way.
- **B2 re-check:** cue grouping, speaker-change flush, SRT/WebVTT labelling and millisecond truncation line up with the JS helper line by line; Rust is safer on the empty inputs where the JS version throws.

**Two things for you**

1. This PR's base is now #178, because it contains #178's commit `76fc7a6e` and both targeted `main` — with squash-merge, whichever landed second would have conflicted.
2. The doc comments state `Segment`/`SentimentSegment`/`TopicSegment::end_word` is *inclusive*. The inference is sound but rests on a single live payload and a token-count proxy, and the Sentiment docs do not say either way. Worth confirming with the API team before we assert it in public rustdoc.

Not fixed, deliberately: the captions example does not pass `utterances`/`diarize`, so the one in-repo demo exercises the plain word-chunking path rather than the B2 fix.

GregHolmes (dismissed):

dg-coreylweathers:
Merged `main` in to clear the conflict — this was blocking the whole stack, since #165–#167 all sit on top of it.

One file conflicted, `CHANGELOG.md`, and only in `## [Unreleased]`: this branch's `### Added` and three `**BREAKING**` bullets against the `**BREAKING**` `send_data`/`finalize`/`keep_alive`/`close_stream` bullet that came in with #179. Kept all four. No code conflicted.

Then cascaded the same merge down the stack so nothing else went conflicting behind it: #164 → #165 → #166 → #167. Only #165 needed a second resolution, where its `cargo doc` bullet is a superset of the one on main (it also covers the shared `common` module's links, which Phase 2's `read` feature introduced) — kept the longer one plus the two streaming bullets from #179.

Gates green on all four after each merge: fmt, clippy, `cargo test --all --all-features`, `cargo doc --workspace --all-features`, all with `-D warnings`.

Heads up @GregHolmes — this dismissed your approval, since the repo drops stale approvals on new commits. Nothing changed but the changelog merge; sorry for the re-request.

GregHolmes (approved):

PR #165, opened 2026-07-31
feat: Phase 2 — Text Intelligence (/read), models endpoints, Stream ergonomics

### LINKED ISSUE #94: Text Intelligence - Analyze (Text to Text)
Feature: [Text Intelligence Docs](https://developers.deepgram.com/docs/text-intelligence)

dg-coreylweathers:
## Phase 2 — Text Intelligence & stream ergonomics

Stacked on top of #164 (Phase 1). Base is `feat/phase-1-quick-wins`, so this diff shows only Phase 2; GitHub will retarget it to `main` once #164 merges.

### What's included
- **#94 — Text Intelligence (`/v1/read`).** `Deepgram::text_intelligence()` → `Read` with `analyze_text` / `analyze_url`, an `Options` builder (sentiment, summarize [boolean, not `v2`], topics, intents, custom topics/intents + modes, language, tag), and a response module reusing the shared sentiment/topic/intent types with a Read-specific `Summary { text }`. Example `analyze_text`.
- **Models-listing endpoints.** `Deepgram::models()` → `Models` with `get_models`, `get_model`, `get_project_models`, `get_project_model`; typed `ModelsResponse` / `Model` / `ModelMetadata`. Example `models`.
- **#163 — `futures::Stream`.** `TranscriptionStream` already implements `Stream<Item = Result<StreamResponse>>`; re-exported at the crate root and documented, with a new `stream_futures` example showing `StreamExt` combinators. Verified against the current code (`src/listen/websocket.rs:793`).

- build / clippy `-D warnings` / fmt / `cargo test --all --all-features` (139 tests, incl. new `/read` + models deserialization tests) ✅
- per-feature `cargo check` (no-default / listen / speak / manage) ✅
- `cargo semver-checks`: **no new breaking changes** — the only 2 breaks are the Phase 1 #130 streaming ones inherited from the base branch.

Closes #94, #163

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Text Intelligence (`/v1/read`) support: `Deepgram::text_intelligence()` returns a `Read` client with `analyze_text` and `analyze_url` for running sentiment, summarization, topic, and intent analysis over text. Reuses the existing sentiment/topic/intent response types and adds a Read-specific `Summary` (`text`) ([#94](https://github.com/deepgram/deepgram-rust-sdk/issues/94)). New example `analyze_text`.
+- Model-listing endpoints on the management API: `Deepgram::models()` returns a `Models` client with `get_models`, `get_model`, `get_project_models`, and `get_project_model`. New example `models`.
+- `futures::Stream` for live transcription is confirmed and documented: `listen::websocket::TranscriptionStream` already implements `Stream<Item = Result<StreamResponse>>`, so results compose with `StreamExt` combinators. New example `stream_futures` demonstrates idiomatic combinator usage ([#163](https://github.com/deepgram/deepgram-rust-sdk/issues/163)).
 - Text-to-Speech REST responses now expose their metadata headers, including the `dg-request-id`. New `Speak::speak_to_file_with_metadata` and `Speak::speak_to_stream_with_metadata` return a `SpeakMetadata` (request id, model name/uuid, character count, content type) alongside the audio ([#89](https://github.com/deepgram/deepgram-rust-sdk/issues/89)). New example `text_to_speech_request_id`.
 - Named Deepgram Whisper Cloud model variants: `Model::Whisper`, `Model::WhisperTiny`, `Model::WhisperBase`, `Model::WhisperSmall`, `Model::WhisperMedium`, and `Model::WhisperLarge` (previously only reachable via `Model::CustomId`) ([#128](https://github.com/deepgram/deepgram-rust-sdk/issues/128)).
 - Additional redaction options: `Redact::Pii`, `Redact::Phi`, and `Redact::AggressiveNumbers` ([#87](https://github.com/deepgram/deepgram-rust-sdk/issues/87)).
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,6 +100,21 @@ path = "examples/transcription/websocket/callback_stream.rs"
 required-features = ["listen"]
 
 [[example]]
+name = "stream_futures"
+path = "examples/transcription/websocket/stream_futures.rs"
+required-features = ["listen"]
+
+[[example]]
+name = "analyze_text"
+path = "examples/read/analyze_text.rs"
+required-features = ["listen"]
+
+[[example]]
+name = "models"
+path = "examples/manage/models.rs"
+required-features = ["manage"]
+
+[[example]]
 name = "microphone_stream"
 path = "examples/transcription/websocket/microphone_stream.rs"
 required-features = ["listen"]
--- a/examples/manage/models.rs
+++ b/examples/manage/models.rs
@@ -1,0 +1,41 @@
+//! List the STT and TTS models available on your Deepgram account.
+//!
+//! Run with:
+//!
+//! ```sh
+//! DEEPGRAM_API_KEY=your-key cargo run --example models --features manage
+//! ```
+
+use std::env;
+
+use deepgram::{Deepgram, DeepgramError};
+
+#[tokio::main]
+async fn main() -> Result<(), DeepgramError> {
+    let deepgram_api_key =
+        env::var("DEEPGRAM_API_KEY").expect("DEEPGRAM_API_KEY environmental variable");
+
+    let dg_client = Deepgram::new(&deepgram_api_key)?;
+
+    let models = dg_client.models().get_models(false).await?;
+
+    println!("STT models:");
+    for model in &models.stt {
+        println!(
+            "  {} ({}) v{}",
+            model.name, model.canonical_name, model.version
+        );
+    }
+
+    println!("TTS models:");
+    for model in &models.tts {
+        let accent = model
+            .metadata
+            .as_ref()
+            .and_then(|m| m.accent.as_deref())
+            .unwrap_or("unknown accent");
+        println!("  {} ({}) — {accent}", model.name, model.canonical_name);
+    }
+
+    Ok(())
+}
--- a/examples/read/analyze_text.rs
+++ b/examples/read/analyze_text.rs
@@ -1,0 +1,54 @@
+//! Analyze plain text with Deepgram's Text Intelligence (`/v1/read`) API.
+//!
+//! Run with:
+//!
+//! ```sh
+//! DEEPGRAM_API_KEY=your-key cargo run --example analyze_text --features listen
+//! ```
+
+use std::env;
+
+use deepgram::{
+    read::options::{Language, Options},
+    Deepgram, DeepgramError,
+};
+
+const TEXT: &str = "Hi. Thank you for calling Premier Phone Services. My name is Beth. \
+    I'm sorry to hear your phone is not working. I would be happy to help you today. \
+    Let me pull up your account so we can get this resolved for you right away.";
+
+#[tokio::main]
+async fn main() -> Result<(), DeepgramError> {
+    let deepgram_api_key =
+        env::var("DEEPGRAM_API_KEY").expect("DEEPGRAM_API_KEY environmental variable");
+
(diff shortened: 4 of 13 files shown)


dg-coreylweathers:
Live-verified against the production API (Rust 1.97 container):
- `analyze_text` → `/v1/read`: sentiment/summary/topics/intents deserialized correctly.
- `models` → `/v1/models`: full STT/TTS list parsed, including TTS `metadata.accent`.
- `stream_futures` → STT WebSocket: `StreamExt::filter_map` extracted final transcripts.

Also addressed the review NIT: `Read`'s inner field is now private (matching `Models`/`Scopes`), dropping the misleading `#[allow(unused)]`.

GregHolmes (changes requested):
# Review: deepgram-rust-sdk #165, feat: Phase 2 - Text Intelligence (/read), models endpoints, Stream ergonomics
Verdict: request-changes

## Intent
Add a typed `/v1/read` Text Intelligence client, management model-listing endpoints, and StreamExt usage documentation. This is intentionally stacked on #164 and is assessed relative to `feat/phase-1-quick-wins`; it must not merge until the parent fixes land and GitHub retargets the PR. Intent is based on the PR, issue #94, issue #163, and the author's live-verification comment; no launch or Slack context is linked.

## Blocking
- [B1] `Cargo.toml:60` and `src/lib.rs:35`, Text Intelligence is tied to the WebSocket-only `listen` feature.

**Title:** [B1] `/v1/read` cannot be enabled independently

**Summary:** The new Text Intelligence API uses ordinary HTTP, but the module, its crate-root export, and its example are gated by `listen`. There is no `read` feature, so `default-features = false, features = ["read"]` fails before compiling and users must pull WebSocket dependencies to call `/v1/read`.

**Expected:** Text Intelligence can be enabled with a dedicated feature without enabling live transcription.

**Observed:** `cargo check --no-default-features --features read` fails with "the package 'deepgram' does not contain this feature: read"; enabling the feature requires `listen` instead.

**Recommended fix:** Add `read = []`, gate `read` and its re-export on that feature, set the example's `required-features = ["read"]`, and make `common` available under either `listen` or `read` because the Read types reuse its options and responses. Verify with `cargo check --no-default-features --features read`.

- [B2] `src/manage/models/response.rs:31`, the typed model response rejects documented valid API responses.

**Title:** [B2] Model fields marked optional by the API are required in Rust

**Summary:** The current Models API reference marks `name`, `canonical_name`, `architecture`, `version`, and `uuid` as optional for both STT and TTS model objects. The new `Model` makes every one a required `String`, so any contract-valid response omitting one field causes deserialization of the entire endpoint response to fail.

**Expected:** A response that matches the documented optional-field schema deserializes successfully and preserves field absence.

**Observed:** The client returns a deserialization error instead of a usable model record whenever one documented optional scalar is omitted.

**Recommended fix:** Model those scalar fields as `Option<String>` (as the feature-specific fields already are), update the example to handle absent display values, and add an omission fixture to the deserialization test.

## Should-fix
- [S1] `src/lib.rs:40`, the Stream re-export claimed by the PR is missing.

**Title:** [S1] Stream ergonomics change does not provide the stated crate-root API

**Summary:** `TranscriptionStream` already publicly implements `futures::Stream` at `listen::websocket::TranscriptionStream`. The PR body says it is re-exported at the crate root, but the only new re-export is `Read`; no source change makes `deepgram::TranscriptionStream` available.

**Expected:** The implementation matches the stated ergonomic crate-root re-export, or the PR accurately describes the existing module-path API.

**Observed:** Consumers must continue importing `deepgram::listen::websocket::TranscriptionStream`.

**Recommended fix:** Add `#[cfg(feature = "listen")] pub use listen::websocket::TranscriptionStream;` to `src/lib.rs`, or remove the re-export claim from the PR description.

## Nits
- None.

## Verified (evidence)
- `/v1/read` uses the documented request body forms (`text` or `url`) and boolean `summarize` query parameter. Its response metadata and summary shapes match the live Text Intelligence documentation.
- Model-listing endpoints, `include_outdated`, response container names, TTS metadata fields, and model endpoint paths match the live Models API reference.
- `TranscriptionStream` already implements `futures::Stream<Item = Result<StreamResponse, DeepgramError>>` at `src/listen/websocket.rs:793`; the added example correctly demonstrates `StreamExt::filter_map`.
- Local Rust 1.94.1: build, tests (72 unit plus 139 doc-tests), clippy, fmt, docs, and existing no-default feature checks pass. The `rust:1.97` image used by the PR contains neither `cargo` nor `rustup`, so it cannot execute the declared container gate.
- The PR is mergeable against its declared parent branch. It remains gated by #164, which has an outstanding requested-changes review and conflicts with current `main`; `git merge-tree --write-tree origin/main HEAD` confirms conflicts in `Cargo.toml` and `CHANGELOG.md`.

## Needs human
- Greg: confirm whether the stack still intends one accumulated 0.11.0 release. The inherited #164 SemVer failures are not new in this diff, but Phase 2 cannot be released or retargeted until #164 is fixed and rebased. This is release sequencing, not a Phase 2 API decision.

## Developer-facing messaging
- The changelog accurately describes the existing module-path `TranscriptionStream` API; the PR body should not promise a crate-root re-export until [S1] is implemented.
- Publishing `/v1/read` behind a `listen` flag makes a text-only client look like it requires streaming transcription. A dedicated `read` feature keeps the install path and product boundary clear.

dg-coreylweathers:
## Round 2 — pushed

Rebased onto the updated #164. One blocking fix, and two of my own findings that live probing reversed.

- **Two shipped skills had a Cargo snippet that does not parse.** `deepgram = { default-features = false, features = ["manage"] }` has no `version`, `path`, `git` or `workspace` key, so cargo refuses the whole manifest: *"dependency (deepgram) specified without providing a local path, Git repository, version, or workspace dependency to use."* These blocks exist to be pasted verbatim by a developer or an agent. Both are now pinned to `0.12`, and I re-ran `cargo metadata` to confirm they parse. Unpinning to avoid teaching a stale version was right; deleting the key was not the way to do it.
- **Three changelog claims are now dated.** "194 entries live", `multilingual` "present on every STT record", and `display_name` sent "for every voice" read as timeless facts but are point-in-time observations. They now say when they were checked (re-verified 2026-09-18).

**What I checked against the live API**

- **Is `language` really required on `/v1/read`?** Yes. Every feature combination without it returns `400 INVALID_QUERY_PARAMETER … missing field 'language'`. The SDK already defaults it to `en` and has a test citing that 400, so this code is correct — but **the API reference is wrong**: it documents `language` as optional with a server-side default of `en`. That is a docs fix, not an SDK fix.
- **Can a sparse analysis object break deserialization?** No. I probed three inputs chosen to detect nothing; `topics` and `intents` came back as `{"segments": []}` and `sentiments` always carried both `segments` and `average`. I had flagged the missing `#[serde(default)]` on the nested arrays as a real failure mode — it is not reachable against the live API, so it is defensive hardening at most.
- **Was relaxing the model fields to `Option` safe?** Yes — `/v1/models` populates `name`, `canonical_name`, `architecture`, `version` and `uuid` in practice, so the relaxation costs nothing and covers the documented-optional case.

**One for a maintainer:** `get_project_models` / `get_project_model` also follow `base_url`, but `src/lib.rs` and the management skill document that exception as applying to `/v1/models` only. A self-hosted client would send those two account-management calls to its own instance. Either narrow the routing or widen the doc.

GregHolmes (changes requested):
**[B1] `src/read/response.rs:108` — Optional Text Intelligence summary text is required by the SDK**

**Summary:** The published `/v1/read` reference marks `results.summary.text` optional, but `Summary::text` is a required `String`. Serde rejects a successful response containing `{"results":{"summary":{}}}` before the caller can inspect any result.

**Expected:** Every response shape the published `/v1/read` contract permits deserializes into the typed SDK response.

**Observed:** A 200 response that omits `results.summary.text` returns a JSON deserialization error instead of `Summary { text: None }`.

**Recommended fix:** Change `Summary::text` to `Option<String>` and add a response fixture proving that an empty `summary` object deserializes. Update call sites and examples to handle `None`.

dg-coreylweathers:
Pushed. `Summary::text` is now `Option<String>`, with the fixture from your repro (`{"metadata":{},"results":{"summary":{}}}` → `Summary { text: None }`) plus a second fixture built from a real live response carrying all four features. `examples/read/analyze_text.rs` and the Text Intelligence skill snippet both go through `.and_then(|s| s.text.as_deref())` now, so nothing unwraps.

One honesty note on the doc comment: I could not get the live endpoint to omit `text`. Six real `/v1/read` calls, including inputs as degenerate as `"a"` and `"..."`, all returned it — the API echoes the input back rather than sending an empty summary. So this is a contract-conformance fix, not a reproduced production failure, and the rustdoc says exactly that rather than implying I saw it break.

No changelog `Changed` line: the `read` feature is itself unreleased, so `Summary::text` has never shipped and this is not a semver event against 0.11.0. Folded into the existing `Added` bullet instead.

**One thing I deliberately did not fix, so it doesn't surprise you in round 4.** The spec also marks every field of the sentiment/topic/intent payloads optional — `SharedSentiments`, `SharedTopics`, `SharedIntents` and their segment types all carry `required: []` — while the SDK requires all of them. I left them alone because those schemas are referenced by `ListenV1ResponseResults` as well as `ReadV1ResponseResults`, so optioning them is a breaking change to the pre-recorded `/v1/listen` public API, not a Phase 2 fix. Two things also make me doubt `required: []` is a deliberate contract statement there: the spec's summary path is nonsense nesting (`results.summary.results.summary.text`) that the live API never produces, and it types `confidence_score` as a string when the API plainly returns a number. Happy to take it as its own PR with its own semver call — I just didn't want to widen a re-review fix into a cross-surface break.

Gates green: fmt, clippy, `cargo test --all --all-features` (172 unit, 146 doc), `cargo doc --workspace --all-features`, all with `-D warnings`.

GregHolmes (approved):

Issue #95, opened 2024-08-28
Text to Speech - Websocket API


DamienDeepgram:
Websocket Streaming (Text to Speech)
- Docs TBD

dg-coreylweathers:
Implemented in #166 (Phase 3): streaming Text-to-Speech over a WebSocket. `Speak::speak_stream()` opens a session (`SpeakStreamHandle`) with `speak`/`flush`/`clear`/`close` and a `futures::Stream` of `SpeakResponse` (audio + Metadata/Flushed/Cleared/Warning). Live-verified against production. Will close when v0.11.0 lands on `main`.

PR #166, opened 2026-07-31
feat: Phase 3 — streaming TTS WebSocket & self-hosted distribution credentials

### LINKED ISSUE #148: [Enhancement] Add Text-to-Speech WebSocket streaming support
## Summary

Add TTS WebSocket streaming support to the Rust SDK, enabling real-time text-to-speech audio generation. This would allow Rust developers to stream text to Deepgram and receive audio chunks in real-time, essential for voice agent and interactive applications.

## Problem it solves

Rust developers building low-latency voice applications (game engines, embedded systems, real-time communication servers) need streaming TTS for interactive speech synthesis. The Rust SDK currently has an open issue (#95) for TTS WebSocket support but no implementation. TTS REST is available but doesn't support streaming, which is required for real-time voice agent use cases where audio needs to start playing before the full response is generated.

## Proposed API

```rust
use deepgram::tts::{TtsWebSocket, TtsOptions, TtsEvent};

let options = TtsOptions::builder()
    .model("aura-2-en")
    .encoding(AudioEncoding::Linear16)
    .sample_rate(16000)
    .build();

let mut tts_ws = deepgram.tts().websocket(options).await?;

// Send text for synthesis
tts_ws.send_text("Hello, how can I help you today?").await?;
tts_ws.flush().await?;

// Receive audio chunks
while let Some(event) = tts_ws.next().await {
    match event? {
        TtsEvent::Audio(data) => { /* play or forward audio bytes */ },
        TtsEvent::Flushed => { /* all audio for current text received */ },
        TtsEvent::Warning(msg) => { /* handle warning */ },
    }
}
```

## Acceptance criteria

- [ ] TTS WebSocket client with streaming text input and audio output
- [ ] Support for flush and close operations
- [ ] Typed event handling (Audio, Flushed, Warning, Close)
- [ ] Documented with usage example
- [ ] Compatible with existing SDK async architecture (tokio)
- [ ] Includes example in `examples/` directory

---
*Raised by the DX intelligence system.*

dg-coreylweathers:
## Phase 3 — Streaming TTS WebSocket + self-hosted credentials

Stacked on #165 (Phase 2). Base is `feat/phase-2-read-stream-models`; GitHub will retarget to `main` as the stack merges.

### What's included
- **#148 / #147 / #95 — Streaming TTS over a WebSocket.** `Speak::speak_stream()` → `SpeakStreamBuilder` (model / encoding / sample rate); `handle()` opens the connection → `SpeakStreamHandle` with `speak` / `flush` / `clear` / `close`, and receives audio + events. `SpeakStreamHandle: futures::Stream<Item = Result<SpeakResponse>>`. `SpeakResponse` (`Audio` / `Metadata` / `Flushed` / `Cleared` / `Warning` / `Unknown`) is `#[non_exhaustive]`; unknown message types are preserved rather than breaking the stream. The `speak` feature now pulls the WebSocket deps. Modeled on `src/listen/websocket.rs`. Example `text_to_speech_websocket`.
- **Self-hosted distribution credentials.** `Deepgram::self_hosted()` → `SelfHosted` with `list` / `get` / `create` / `delete_distribution_credentials` + typed response models. Example `self_hosted_credentials`.

- build / clippy `-D warnings` / fmt / `cargo test --all --all-features` (142 tests, incl. TTS-WS protocol + self-hosted deserialization tests) ✅
- per-feature `cargo check` incl. **speak-only** (now pulls WS deps) ✅
- `cargo semver-checks`: **no new breaking changes** (only the inherited Phase 1 #130 streaming breaks).
- **Live-verified against production:** TTS WebSocket connected, emitted `Metadata` + `Flushed`, and streamed 219 KB of audio. Self-hosted list returns a correct `403 INSUFFICIENT_PERMISSIONS` on accounts without the self-hosted scope (confirms the request path).

Closes #148, #147, #95


dg-coreylweathers:
Addressed the review's BLOCKING finding: the TTS-WS worker busy-spun (and hung on current-thread runtimes) after input close because it kept selecting on the drained input channel. `run_worker` now stops selecting on `message_rx` once input is closed and only awaits server messages (via a shared `handle_incoming` helper). **Verified**: re-ran the example under `#[tokio::main(flavor = "current_thread")]` — it completes cleanly (219KB audio) where it previously would hang. Also documented the send-then-drain contract and the nil-`request_id` fallback on `SpeakStreamHandle`. The self-hosted create/get envelope matches `list` per the API docs (no secret token field in the create response).

GregHolmes (changes requested):
# Review: deepgram-rust-sdk #166, feat: Phase 3 - streaming TTS WebSocket and self-hosted distribution credentials
Verdict: request-changes

## Intent
Add WebSocket streaming TTS and self-hosted distribution-credential management while keeping the SDK's existing async patterns. This PR is intentionally stacked on #165 and cannot merge until #164 and #165 are fixed, merged, and the stack is retargeted. Intent is based on the PR, issues #95, #147, and #148, and the author's live-verification comment; no launch or Slack context is linked.

## Blocking
- [B1] `src/speak/websocket.rs:161`, streaming TTS silently drops documented metadata.

**Title:** [B1] `SpeakResponse::Metadata` omits `additional_model_uuids`

**Summary:** The streaming TTS API's `Metadata` event includes the optional `additional_model_uuids` array, but neither `TextEvent::Metadata` nor the public `SpeakResponse::Metadata` carries it. Serde ignores the unknown field, so a successful request silently loses model provenance that a caller may need for auditing or billing.

**Expected:** All documented fields from a streaming TTS Metadata event are available in the typed response.

**Observed:** `additional_model_uuids` is discarded during deserialization and cannot be recovered by SDK users.

**Recommended fix:** Add `additional_model_uuids: Option<Vec<Uuid>>` to `TextEvent::Metadata` and `SpeakResponse::Metadata`, pass it through the conversion, and add a fixture that asserts the values survive parsing.

- [B2] `src/speak/websocket.rs:98`, the streaming TTS builder does not enforce or expose the endpoint's option contract.

**Title:** [B2] Streaming TTS accepts invalid encodings and omits supported controls

**Summary:** The streaming endpoint supports only `linear16`, `mulaw`, and `alaw`, but `SpeakStreamBuilder::encoding` accepts the REST `Encoding` enum, including `mp3`, `opus`, `flac`, `aac`, and arbitrary custom values. Conversely, the endpoint documents `speed` and `mip_opt_out`, but the builder has no setters or query-parameter escape hatch, so callers cannot use them at all.

**Expected:** The streaming API rejects locally known-invalid encodings before connecting and exposes every documented streaming control.

**Observed:** `Encoding::Mp3` is serialized into a request the server rejects, while `speed` and `mip_opt_out` cannot be sent through this client.

**Recommended fix:** Use a streaming-specific encoding enum or validate the existing enum in `handle()` with a clear `DeepgramError::InvalidOptions`; add `speed` and `mip_opt_out` to the builder, with the documented `0.7..=1.5` range for `speed`, and cover valid and invalid URLs in tests.


## Should-fix
- [S1] `src/speak/websocket.rs:534`, protocol tests do not cover every supported client and server message.

**Title:** [S1] Streaming TTS protocol coverage omits Clear, Cleared, and Warning paths

**Summary:** The unit tests assert Speak, Flush, Close, Metadata, Flushed, and Unknown behavior, but omit Clear transmission and Cleared and Warning parsing. These are public control paths; a future serde rename or refactor can break them without a test failure.

**Expected:** Each public protocol message has an exact serialization or parsing assertion.

**Observed:** The available tests do not exercise Clear, Cleared, or Warning.

**Recommended fix:** Extend the existing table-style tests with those three cases, including optional `sequence_id`, `description`, and `code` values.

## Nits
- None.

## Verified (evidence)
- The streaming TTS WebSocket path, ClientMessage variants, required metadata fields, Flush/Clear/Close messages, and supported encoding list align with the live API reference.
- The self-hosted list and create paths, request body/query placement, defaults, and response envelope align with the live Self-Hosted API reference.
- Local Rust 1.94.1: build, tests (79 unit plus 140 doc-tests), clippy, fmt, docs, and `cargo check --no-default-features --features speak` pass. The `rust:1.97` image named in the PR contains neither `cargo` nor `rustup`, so it cannot run the declared container gate.
- The prior busy-spin report is fixed in the current worker: once the input closes, it stops selecting on the drained input channel and drains server messages only.

## Needs human
- Greg: the stack depends on #164 and #165. Both have requested-changes reviews and #164 conflicts with `main`, so this PR must stay unmerged until their fixes are merged and the stack is retargeted. The inherited SemVer failures are a release-process decision for the planned 0.11.0 aggregation.

## Developer-facing messaging
- The changelog accurately says the `speak` feature now pulls WebSocket dependencies. It should not imply the typed `Metadata` event is complete until [B1] preserves `additional_model_uuids`.
- Rejecting unsupported output encodings before opening the socket and exposing documented controls gives developers a clear local error instead of an avoidable server-side failure.

dg-coreylweathers:
## Round 2 — pushed, and the self-hosted question is answered

Rebased onto the updated #165. Your three items are fixed; I also probed the self-hosted contract live, because three claims in the code contradicted the published reference.

**The self-hosted scope concern turned out to be a docs bug, not an SDK bug**

The reference documents `scopes` and `provider` as **query** parameters on `create`, while this SDK sends both in the JSON body. If the reference had been right, a caller narrowing the grant with `.scopes([Scope::Api])` would have silently received the default `self-hosted:products` — every product image — with no error. I tested both forms against production:

- The documented **query** form returns `400 Json deserialize error: missing field 'provider'`. The server reads `provider` from the body.
- The SDK's **body** form reaches permission checking and returns `403 INSUFFICIENT_PERMISSIONS` whose details quote back the scope supplied *in the body*: "Check that your account has the 'self-hosted:product:api' scope for this project."

So the server reads both from the body, the narrowed scope is received, and the SDK is correct. **The reference page and the OpenAPI spec are wrong** — routing that to @GregHolmes separately. No credential was created (403), so nothing was provisioned.

- Skill manifest pinned to `0.12` — it had lost its `version` key, which makes the snippet unparseable by cargo.
- "covering every voice the API serves" is now scoped to this release, in both the changelog and the skill. The exact-count test still enforces the list, and its comment already records the live probe behind the one exclusion (`aura-2-perseo-it`).

**What I verified**

- B1: `additional_model_uuids` is `Option<Vec<String>>` with `#[serde(default)]`, absent and empty both covered. Confirmed live on the REST side too, where the same data arrives as `dg-additional-model-uuids` — and was being dropped there (fixed in #164).
- B2: `validate()` rejects `mp3`/`opus`/`flac`/`aac` with `InvalidOptions` *before* opening a socket, and the `0.7..=1.5` speed range matches the documented `speed_out_of_range` error.
- Lifecycle: capacity reserved before inbound reads, terminal write error delivered exactly once through a dedicated reserved slot, no busy-spin after input close, worker exits without waiting for handles. The backpressure test overflows both bounded channels at once and still delivers everything.
- Self-hosted list works live (HTTP 200) on both the `self-hosted` and legacy `onprem` paths.

**Still open:** the `create` response shape. This project's key lacks the `self-hosted:product:*` scope, so a 200 create is not reachable from here and the flat-vs-wrapper question stays author-asserted. Needs one create on an account carrying the scope.

GregHolmes (changes requested):
**[B1] `src/manage/self_hosted.rs:293` — Successful credential creation can be reported as a client error**

**Summary:** `create_distribution_credentials` deserializes a bare `CreatedDistributionCredentials` object with one-time `username` and `secret`. The published endpoint reference instead documents a `{ member, distribution_credentials }` response, and the PR records that no scoped account was available to observe a successful server response. If the published response is what the server returns, creation succeeds remotely but the SDK returns a JSON error; a caller that retries may create duplicate credentials while never receiving the original result.

**Expected:** The SDK create-response type matches a verified successful server response and returns it without a deserialization failure.

**Observed:** The response model is author-asserted rather than exercised, and it conflicts with the published success schema.

**Recommended fix:** Run one create request against an account with the required self-hosted scope, capture the sanitized 200 response, and model that shape. If it is the documented wrapper, return `DistributionCredentialsEntry`; if it is the bare secret-bearing object, retain the current type and open a docs/spec correction with the captured evidence.

dg-coreylweathers:
I can't close B1 from this account, and I don't want to guess the shape — that's exactly what you flagged. Here's what I established and what I need.

**Blocked on scope.** Our member is `["owner"]` on both prod and staging, and owner is not sufficient. Every create attempt, both hosts:

```
{"category":"INSUFFICIENT_PERMISSIONS","message":"Your account does not have the required scope to perform that action for this project.","details":"Check that your account has the 'self-hosted:product:api' scope for this project."}
```

Tried all seven `self-hosted:product:*` values. The list endpoint returns `{"distribution_credentials":[]}` on both, so it constrains nothing either. **I never observed a successful create response, so I changed nothing about `CreatedDistributionCredentials`.**

What would unblock it: a project on staging whose member holds any one of `self-hosted:product:api` / `engine` / `license-proxy` / `dgtools` / `billing` / `hotpepper` / `metrics-server`. One create plus one delete settles it in a minute; the probe script already does create → GET → list → delete → verify-zero.

**What I did verify — and the reference is wrong about it.** The create request contract:

- `provider`, `comment` and `scopes` are all **required JSON body** fields. Omit any one and you get `Json deserialize error: missing field \`<name>\``, including when you supply it as a query parameter instead.
- Query parameters are not read at all.
- `provider` accepts only `quay` — anything else gives `unknown variant \`bogus-provider\`, expected \`quay\``.

The reference documents `scopes` and `provider` as *query parameters with defaults* and `comment` as an *optional* body field. All three are wrong. The PR's implementation already matches the server; the only change here turns an author-asserted test comment into the verified contract, with the date.

Two more for the spec, separate from this PR:
1. The create 200 never mentions `username` or `secret`, and reuses the GET-by-id response description verbatim. Whichever model turns out to be right, the reference can't currently be the source for either.
2. Our own `api` skill lists the domain as `/v1/projects/*/selfhosted/*`, which 404s. Live paths are `self-hosted` and `onprem` (both 200).

**A caution for whoever runs the scoped create.** I nearly shipped a wrong fix here. `scopes: ["self-hosted:products"]` returns a 400 identical to a bogus string, which reads like "the shorthand is invalid" — I had a whole change written removing `DEFAULT_SCOPE`. It isn't invalid: the roles guide says the shorthand expands to the scopes the caller actually holds, and ours holds none, so it expands to empty. From an unentitled account those two hypotheses are indistinguishable. Re-run that probe from the scoped account; it'll settle whether `DEFAULT_SCOPE` is usable.

GregHolmes (changes requested):
**[B1] Self-hosted credential creation has no verified success-response contract**

**Summary:** `SelfHosted::create_distribution_credentials` decodes only a bare `CreatedDistributionCredentials` object with one-time `username` and `secret`, while the published endpoint reference documents a `{ member, distribution_credentials }` wrapper. The available account cannot obtain a successful create response, so the bare shape remains unverified. If production returns the documented wrapper, the credential is created but the SDK returns a deserialization error; a retry can create duplicates without exposing the original secret.

**Expected:** The return type and example are based on one sanitized HTTP 200 create response from an entitled project.

**Observed:** `src/manage/self_hosted/response.rs` and its fixtures encode the bare secret-bearing shape without a captured successful create response, despite the conflicting public schema.

**Recommended fix:** Run create, get, list, and delete against a project whose member has a `self-hosted:product:*` scope. Model the captured 200 response and add it as a fixture. If it is the documented wrapper, return `DistributionCredentialsEntry`; if it is the bare secret-bearing shape, retain the current type and file a `deepgram-docs` / OpenAPI correction with the evidence.

PR #167, opened 2026-07-31
feat(agent): Phase 4 — Voice Agent WebSocket client (#152)

### LINKED ISSUE #152: [Enhancement] Add Voice Agent WebSocket client support
## Summary

Add a Voice Agent WebSocket client to the Rust SDK, enabling developers to build voice agent applications in Rust. This would provide typed message structures for the Voice Agent API's WebSocket protocol and an async client for managing agent sessions.

## Problem it solves

Rust developers building voice agent applications currently have no SDK support for Deepgram's Voice Agent API. They must manually implement the WebSocket protocol, message serialization, and connection lifecycle. The Rust SDK already supports STT (pre-recorded and streaming) and has TTS WebSocket streaming in progress (#148), but Voice Agent — Deepgram's highest-growth API surface — is missing entirely. This blocks Rust adoption for real-time voice AI applications, embedded systems, and high-performance server-side agent orchestration.

## Proposed API

```rust
use deepgram::agent::{AgentClient, AgentConfig, AgentEvent};

let config = AgentConfig::builder()
    .model("nova-3")
    .voice("aura-asteria-en")
    .language("en")
    .build();

let mut agent = client.agent().connect(config).await?;

// Send audio frames
agent.send_audio(&audio_bytes).await?;

// Receive events
while let Some(event) = agent.next_event().await? {
    match event {
        AgentEvent::Transcript { text, is_final } => { /* ... */ }
        AgentEvent::Audio(bytes) => { /* play TTS audio */ }
        AgentEvent::FunctionCall { name, args } => { /* handle tool call */ }
        AgentEvent::Error(e) => { /* handle error */ }
    }
}
```

## Acceptance criteria

- [ ] Typed request/response message structures for Voice Agent WebSocket protocol
- [ ] Async WebSocket client with connection lifecycle management (connect, keepalive, disconnect)
- [ ] Support for audio input streaming and TTS audio output
- [ ] Function calling / tool use message handling
- [ ] InjectAgentMessage and UpdatePrompt support
- [ ] Documented with usage example
- [ ] Compatible with existing SDK architecture and async runtime (tokio)

---
*Raised by the DX intelligence system.*

dg-coreylweathers:
## Phase 4 — Voice Agent WebSocket client

Stacked on #166 (Phase 3). Base is `feat/phase-3-tts-ws-selfhosted`; GitHub retargets to `main` as the stack merges. This is the strategic bet from the parity assessment.

### What's included
- New `agent` feature + module. `Deepgram::agent()` → `Agent`; `start()` / `start_at_url()` open a live session at `wss://agent.deepgram.com/v1/agent/converse`, returning an `AgentHandle` (send `Settings`, `UpdateSpeak`/`UpdateThink`/`UpdatePrompt`, `InjectUserMessage`/`InjectAgentMessage`, `FunctionCallResponse`, `KeepAlive`, and binary audio) and an `AgentEventStream` (`futures::Stream`) of typed `AgentResponse` events (`Welcome`, `SettingsApplied`, `ConversationText`, `UserStartedSpeaking`, `AgentThinking`, `FunctionCallRequest`, `AgentStartedSpeaking`, `AgentAudioDone`, `History`, `Error`, `Warning`, mid-session `*Updated` acks, `Unknown`).
- Fully typed `Settings`: inline or saved (`Uuid`) configs; `listen`/`think`/`speak` provider settings (Deepgram STT V1/V2; OpenAI/Anthropic/AWS Bedrock/Google/Groq think; Deepgram/ElevenLabs/Cartesia/OpenAI/AWS Polly speak); audio I/O; conversation history. All event/response types `#[non_exhaustive]`; unknown messages preserved.
- Examples `simple_agent`, `function_calling`.

### Provenance & scope
Ported and adapted from the maintainer's `feat/agent-websocket` branch — the **live WebSocket client only**. The REST saved-configuration / variables / think-models namespace is **deferred** per the assessment ("deferred until the live Voice Agent client exists").

- build / clippy `-D warnings` / fmt / `cargo test --all --all-features` (207 lib + 141 doctests) ✅
- per-feature `cargo check` incl. **agent-only** ✅; doc build `-D warnings` ✅
- `cargo semver-checks`: **no new breaking changes** (only the inherited Phase 1 #130 breaks).
- **Live-verified end-to-end against production:** connected, `Welcome`, `Settings applied`, received greeting audio + `ConversationText` "Hello! How can I help today?" + `History` + `AgentAudioDone`. Also verified the session runs cleanly under a **current_thread** runtime (no hang).

### Note (fix beyond the port)
The ported agent worker had the same latent busy-spin/current-thread-hang as the Phase 3 TTS-WS worker (kept selecting on the drained input channel after close). Fixed here identically (stop selecting on `message_rx` once input closes; shared `handle_agent_inbound` helper) and verified under `current_thread`.

Closes #152


dg-coreylweathers:
Review addressed. Worker fix verified correct by the review (no busy-spin, no dropped messages, ordering/Ping/close semantics preserved; verified live under `current_thread`). Fixed the one SHOULD-FIX: four `[bracket]` intra-doc links pointed at `listen`/`speak`-gated items and broke `cargo doc --no-default-features --features agent -D warnings`; demoted them to plain code-spans (agent-only doc now builds clean). Also documented the send-then-drain backpressure contract on `AgentHandle`. The untagged-`AgentResponse` degrade-to-`Unknown` and shared bounded-channel caveats are intentional/forward-compat and consistent with the existing Flux client.

GregHolmes (changes requested):
# Review: deepgram-rust-sdk #167, feat(agent): Phase 4 - Voice Agent WebSocket client (#152)
Verdict: request-changes

## Intent
Add the live Voice Agent WebSocket client with typed settings, messages, events, function calling, binary audio, and examples. REST saved-configurations remain deliberately deferred. This PR is stacked on #166 and must not merge until #164 through #166 are fixed, merged, and the stack is retargeted. Intent is based on the PR, issue #152, the API reference, and the author's response; no launch or Slack context is linked.

## Blocking
- [B1] `src/agent/listen.rs:191`, Flux language hints serialize under the wrong key and shape.

**Title:** [B1] Flux language prompting is silently dropped

**Summary:** The Voice Agent V2 listen contract uses `language_hints` as an array, but the SDK emits `language_hint` and serializes one hint as a scalar. It also omits the V2 `eot_threshold`, `eager_eot_threshold`, and `eot_timeout_ms` fields, so a caller cannot configure the documented Flux turn-detection behavior.

**Expected:** A V2 listen configuration emits the documented `language_hints` array and exposes the documented turn-detection settings.

**Observed:** The server receives an unrecognized language-hint field, and the supported V2 controls have no typed SDK path.

**Recommended fix:** Rename the field and builder to `language_hints`, serialize it as `Vec<String>` in every case, add the three EOT fields and builder methods with documented ranges, and assert the exact JSON for single and multiple language hints.

- [B2] `src/agent/messages.rs:45` and `src/agent/websocket.rs:237`, supported session controls cannot be sent.

**Title:** [B2] AgentHandle omits `UpdateListen` and `ForceEndTurn`

**Summary:** The API supports `UpdateListen` and `ForceEndTurn`, but there is neither a typed message nor an AgentHandle method for either one. This is not recoverable through the SDK because `send_json` is private; notably, users cannot end a Flux turn when `eot_threshold = 1.0` suppresses natural endpointing.

**Expected:** Every documented live-session control has a typed sender or a documented raw-message escape hatch.

**Observed:** Callers cannot update the listen provider or force a turn to end through this client.

**Recommended fix:** Add `UpdateListenMessage` and `ForceEndTurnMessage`, expose matching `send_update_listen` and `force_end_turn` methods, and test their wire JSON.

- [B3] `src/agent/response.rs:38` and `src/agent/think/mod.rs:120`, function cancellation and safe deferral are unavailable.

**Title:** [B3] Function callers cannot prevent invalidated side effects

**Summary:** `FunctionCallCancelled` is delivered as untyped `Unknown`, even though the server explicitly tells clients not to respond to its cancelled IDs. `ThinkFunction` also omits `defer_until_eot`, so clients cannot defer an irreversible function until the user turn is confirmed; a user interruption can therefore leave an application executing an invalidated action.

**Expected:** A function-calling client receives typed cancellation IDs and can configure documented defer-until-end-of-turn behavior.

**Observed:** The cancellation signal is not actionable without manually parsing raw JSON, and safe deferral cannot be sent.

**Recommended fix:** Add a typed `FunctionCallCancelled` response and `defer_until_eot: Option<bool>` to `ThinkFunction`, with tests covering cancellation and an interrupted/deferred function configuration.

- [B4] `src/agent/speak/deepgram.rs:39`, advertised Deepgram V2/Flux speech settings cannot be emitted.

**Title:** [B4] Deepgram Speak provider is limited to V1

**Summary:** The PR advertises Deepgram V1 and V2 provider settings, but `DeepgramSpeakVersion` contains only `V1`. A caller using `DeepgramSpeakModel::Other("flux-...")` still cannot set `version: "v2"`; the typed model list also has no Flux voices and lacks Flux expressivity.

**Expected:** The advertised Deepgram V2/Flux Voice Agent configuration serializes its version and supported settings.

**Observed:** All Deepgram speak requests are V1-shaped, so Flux Voice Agent TTS cannot be configured through the typed API.

**Recommended fix:** Add V2, Flux model values, and the V2-specific configuration fields from the Voice Agent schema, then add a V2 serialization fixture.

- [B5] `src/agent/websocket.rs:376`, receive-first selection can indefinitely starve outbound work.

**Title:** [B5] Continuous inbound audio can starve microphone audio and controls

**Summary:** `select_biased!` always selects `ws_recv` before the outbound channel. While agent audio is continuously readable, outgoing binary audio, function responses, KeepAlive messages, and close requests may never be sent; the bounded sender eventually blocks despite the documented send-and-drain pattern.

**Expected:** Incoming and outgoing WebSocket work make progress concurrently during a live session.

**Observed:** A continuously ready receive stream can prevent every outbound message from being processed.

**Recommended fix:** Use unbiased selection or an explicit alternating/fair scheduling strategy, and add a localhost WebSocket regression test that continuously sends inbound audio while asserting outbound controls are delivered.

- [B6] `src/agent/websocket.rs:419`, handles can successfully send into an already-dead session.

**Title:** [B6] Terminal worker shutdown silently discards future sends

**Summary:** After a peer close or dropped event consumer, the worker closes only `response_tx` and then waits indefinitely for all `message_rx` senders to disappear. A retained AgentHandle can still report a successful send while the worker discards it, leaving both a misleading success path and a retained task.

**Expected:** After terminal connection state, all AgentHandle send methods fail promptly and the worker exits.

**Observed:** Commands can be accepted but never reach the WebSocket, and the task remains alive until every handle is dropped.

**Recommended fix:** Close the command receiver on every terminal path and return immediately after cleanup; add a test that a retained handle fails sends after server close and after event-stream drop.

- [B7] `src/agent/aws_credentials.rs:13` and `src/agent/endpoint.rs:15`, Debug output exposes secrets.

**Title:** [B7] Voice Agent configuration logs AWS and endpoint credentials

**Summary:** `AwsCredentials`, `Endpoint`, and `FunctionEndpoint` derive `Debug`, recursively exposing AWS access keys, secret keys, session tokens, and custom headers such as `Authorization` when a Settings, Think, or Speak configuration is logged. This contradicts the SDK's existing redacted API-key handling and can leak credentials to application logs.

**Expected:** Debug formatting never reveals secrets or authorization header values.

**Observed:** Standard Rust debug logging prints credential-bearing fields verbatim.

**Recommended fix:** Replace derived Debug with redacted implementations for credentials and header maps, and add tests proving representative secret values are absent from `format!("{value:?}")`.

- [B8] `src/agent/websocket.rs:124`, custom cleartext URLs receive the API credential.

**Title:** [B8] `start_at_url` can transmit credentials over arbitrary `ws://` connections

**Summary:** `start_at_url` accepts any parseable URL and unconditionally adds the Deepgram authorization header. A mistaken non-loopback `ws://` URL sends the API key or temporary token in cleartext; the method explicitly advertises custom self-hosted endpoints, making this a realistic configuration risk.

**Expected:** Credentials are sent only over TLS, except for an explicit safe local-test exception.

**Observed:** Any remote cleartext WebSocket URL receives the authorization header.

**Recommended fix:** Reject non-TLS URLs unless the host is loopback, or require an explicit opt-in that documents the credential risk; add tests for allowed `wss://`, allowed local `ws://`, and rejected remote `ws://` URLs.

## Should-fix
- [S1] `src/agent/websocket.rs:136`, custom handshakes build an incorrect Host authority.

**Title:** [S1] `start_at_url` omits non-default ports from the Host header

**Summary:** `url.host_str()` removes the port and does not create bracketed IPv6 authorities. The documented localhost integration URL therefore sends `Host: 127.0.0.1` rather than `Host: 127.0.0.1:NNNN`; strict proxies and virtual-host routing can reject or misroute it.

**Expected:** The Host header is the complete URL authority, including a non-default port and valid IPv6 formatting.

**Observed:** Custom server handshakes lose port and IPv6 authority information.

**Recommended fix:** Build Host from `url.authority()` or equivalent URL host-and-port formatting, and add localhost-port and IPv6 handshake tests.

- [S2] `src/agent/response.rs:38` and `src/agent/messages.rs:294`, several documented typed surfaces remain unavailable.

**Title:** [S2] Typed API omits ListenUpdated, LatencyReport, and injection interruption behavior

**Summary:** The response union lacks typed `ListenUpdated` and `LatencyReport` variants, and `InjectAgentBehavior` lacks the documented `interrupt` behavior. They degrade to Unknown or cannot be requested, despite the PR's claim that mid-session acknowledgements and responses are fully typed.

**Expected:** The advertised typed event and injection-control surface covers the documented protocol.

**Observed:** Callers must parse raw JSON for two events and cannot request an interruption.

**Recommended fix:** Add the two response types and `InjectAgentBehavior::Interrupt`, with exact serialization and deserialization fixtures.

## Nits
- None.

## Verified (evidence)
- The endpoint, Token authorization scheme, Settings-first ordering, binary audio frames, and documented client/server message families match the live Voice Agent API reference.
- The worker no longer busy-spins after its input channel closes; its input-closed branch drains only inbound server messages.
- Local Rust 1.94.1: build, tests (207 unit plus 141 doc-tests), clippy, fmt, docs, and `cargo check --no-default-features --features agent` pass. The declared `rust:1.97` image lacks `cargo` and `rustup`, so it cannot run the stated container gate.
- The PR is mergeable against its parent but `git merge-tree --write-tree origin/main HEAD` confirms conflicts in `Cargo.toml`, `CHANGELOG.md`, and `src/lib.rs` against current main.

## Needs human
- Greg: confirm whether the accumulated 0.11.0 release plan remains correct after resolving #164 through #166. This PR inherits their requested changes and current-base conflicts, so it cannot receive a merge-go until the stack has been repaired and retargeted.

## Developer-facing messaging
- The changelog's "fully typed" Voice Agent story is currently too broad: several documented controls and events are missing or silently fall through to Unknown. Correcting these gaps is necessary before presenting the SDK as a complete live Voice Agent client.
- The `start_at_url` documentation should state the TLS requirement and local-test exception once [B8] is fixed, so self-hosted users do not accidentally expose their Deepgram credentials.

dg-coreylweathers:
## Round 2 — pushed. Two of your eight were only partly fixed

Rebased onto the updated #166. Re-verifying your items found that B5, B6 and B7 had each been addressed in a way that left the original failure reachable by a slightly different route. Full gate is green: 24/24 steps in Docker, including doctests and every per-feature check.

**B5 — outbound starvation was still reproducible**

The fix swapped `select_biased!` for fair `select!`, but the inbound branch then awaited `response_tx.send(..)` *inside* that branch. Once the consumer fell behind the 256-slot event channel, the worker parked there and never polled the command channel again — so nothing outbound reached the wire: microphone audio, `FunctionCallResponse`, `KeepAlive`, and `close()` alike. Fair selection cannot help, because the branch that wins holds the loop. I reproduced it with a consumer that retains the event stream without reading it: `keep_alive()` returned `Ok` and the server saw nothing.

The worker now reserves event-channel capacity *before* reading a frame and forwards with `start_send`, so a lagging consumer becomes ordinary socket backpressure. That is the structure the streaming TTS worker in #166 already used.

The new regression test discriminates: it **fails** against the previous worker (`KeepAlive starved by an undrained event stream`) and passes against this one. The existing fairness test passed against both — its own comment said so, which is why this stayed hidden.

**B6 — the same stall could beat the shutdown path**

The terminal error forwards awaited that same possibly-full channel, so with a stalled consumer the worker parked *before* `message_rx.close()` ran, and a retained handle kept reporting success into a channel nobody drains. They now use a dedicated sender clone with its own guaranteed slot.

**B7 — `Debug` still printed credentials carried in the query string**

Userinfo and header values were redacted, but the query was printed verbatim — and `think.endpoint` / `speak.endpoint` are exactly where a caller points at Google AI Studio (`?key=`) or Azure OpenAI (`?api-key=`). A partial redaction that reads as complete is worse than none. The value of any query parameter naming a credential is now redacted; non-secret parameters still print, so the endpoint stays identifiable.

**Two more found in this pass**

- **`AwsCredentials` could not be constructed by a consumer.** It is `#[non_exhaustive]` with no constructor and no `Default`, so a struct literal does not compile downstream, and `AwsPollySpeakProvider::new` takes one by value — making the advertised AWS Polly provider unreachable outside this crate and Bedrock's `credentials` permanently `None`. The in-crate test could not catch it, because a struct literal is legal inside the defining crate. Added `iam()`, `sts()`, `new()` and `with_*`.
- **Minimal-Versions was red.** `Receiver::try_recv` arrived in futures 0.3.32 while the floor said `^0.3`. Floor raised to match the API actually used. (`try_next` would lower it, but it is deprecated, so `-D warnings` rejects it — noted at the call site so nobody trades one red job for the other.)

**What I checked against the live agent socket** — the local API reference is name-level only and omits `ForceEndTurn` entirely, so this needed real calls:

- `language_hints` parses as an array; the scalar `language_hint` is rejected outright. Your B1 rename was right.
- `language_hints` requires `flux-general-multi` and is mutually exclusive with `agent.language`. The SDK is safe: `language` is `Option`, defaults to `None`, skips serializing when unset, and is `#[deprecated]` — so it never forces the conflicting combination.
- `ForceEndTurn` is real: accepted on a Flux v2 session, and on a v1 session the server answers `Warning` / `FORCE_END_TURN_UNSUPPORTED`. A bogus message returns `UNPARSABLE_CLIENT_MESSAGE` and drops the connection, so that acceptance means something.
- `UpdateListen` answers `ListenUpdated`; speak `version: v2` with a Flux voice applies; a `defer_until_eot` function applies; `InjectAgentMessage` with `behavior: interrupt` interrupts and the agent speaks.

The changelog claimed audio "cannot starve" the outbound direction — false in exactly the case above. It now states the mechanism and the close-handshake precondition that genuinely holds, names `AgentEvent` as the stream item (it never said how the agent's audio arrives), adds the `send_raw_json` escape hatch, and corrects `speed` to both provider versions with `expressivity` on v2 only.

**Left open, as design calls:** a decode error and a transport error are the same `Err` shape with different consequences, and both examples `break` on either, so one malformed payload ends a live session in our own sample code. There is also no liveness bound on a silent peer, and the escape hatch is whole-message granular. My recommendation is to take the first before this client gets wide adoption, since it changes the error contract.

dg-coreylweathers:
Heads-up on a small conflict this PR will pick up, so it is not a surprise at merge time.

#182 is a 4-line fix on `main` to `.agents/skills/deepgram-rust-voice-agent/SKILL.md`: its `description` contains `Route honestly: this crate...`, and the unquoted `: ` makes the entire YAML frontmatter fail to parse, so the skill loads with no `name` and no `description`. It is the only one of the seven manifests on `main` that fails. That fix is being landed now rather than waiting for this stack.

This PR rewrites that whole file, so the two conflict on exactly that path and nothing else. **Resolve by taking this PR's version wholesale** — the replacement description here has no colon in it, so it parses cleanly (verified), and the "not currently exposed" wording that #182 is repairing is text this PR deletes anyway.

No action needed until #182 merges.

GregHolmes (changes requested):
**[B1] `src/agent/response.rs:404` — Typed latency events fail when the server sends string values**

**Summary:** The published Voice Agent AsyncAPI declares every `AgentStartedSpeaking` and `LatencyReport` latency property as a JSON string with a float title. The SDK models them as `f64` and `Option<f64>`, which rejects valid messages such as `{"type":"LatencyReport","total_latency":"0.83"}`.

**Expected:** String-encoded latency values deserialize into typed latency values.

**Observed:** A valid latency event becomes a JSON deserialization error.

**Recommended fix:** Add a string-or-number float deserializer for every latency field and fixtures for both representations.

**[B2] `src/agent/speak/deepgram.rs:149` — Flux expressivity serializes to the wrong JSON type**

**Summary:** The AsyncAPI defines Flux TTS expressivity as the string enum `"-2"` through `"2"`. The SDK serializes it as an integer and only accepts numeric input.

**Expected:** Flux speak settings use the documented string enum.

**Observed:** The SDK writes `"expressivity":2` rather than `"expressivity":"2"`.

**Recommended fix:** Serialize and deserialize its string representation; add exact Settings and UpdateSpeak fixtures.

**[B3] `src/agent/endpoint.rs:150` — URL-encoded endpoint keys bypass Debug redaction**

**Summary:** Endpoint Debug redaction compares raw query names before URL decoding. Encoded credential names such as `api%2Dkey` or `%6Bey` bypass the redaction list and expose their values in logs.

**Expected:** Debug output redacts credentials regardless of percent-encoding in the parameter name.

**Observed:** `https://example.test/?api%2Dkey=secret` prints `secret` in Debug output.

**Recommended fix:** Percent-decode query names before calling `is_secret_param`, preserve the original encoded name when rendering, and add encoded-key regression tests.

GregHolmes:
Correction to B2: withdrawn. The Python SDK types Voice Agent `expressivity` as `pydantic.StrictInt`, rejects string values such as `"2"`, and its wire tests assert JSON `expressivity: 2`. Rust's numeric encoding matches the cross-SDK contract; the AsyncAPI string schema is stale or incorrect.

dg-coreylweathers:
Pushed three commits. Two of your findings held and I fixed them. **B2 does not hold — implementing it as written would break every Flux TTS voice agent**, and I have live evidence. Details below.

### B3 — percent-encoded credential names bypass Debug redaction
Confirmed and fixed. Parameter names are now matched raw, then after each round of percent-decoding (three rounds, hex digits either case), so `api%2Dkey`, `api%2dkey`, `%6Bey`, `Api%5Fkey`, the doubly-encoded `api%252Dkey` and a fully encoded `%61%70%69%2d%6b%65%79` all redact. The name is reprinted exactly as written so the output still says which parameter was dropped; over-decoding can only over-redact, which is the safe direction for a `Debug` impl.

While in there I found a second leak you didn't flag: the fragment was never scanned. `redact_query_secrets` split on `?` first, so a `?` inside a fragment started a bogus query, and an OAuth implicit-flow `#access_token=…` printed in full. It now splits the fragment off first (RFC 3986) and runs the same scan over both. Four new fixtures, including one asserting a percent-encoded *non*-secret is still printed verbatim and that a trailing `%` doesn't panic.

### B1 — string-encoded latency values
The spec backs you: every latency property on `AgentStartedSpeaking` and `LatencyReport` is `type: string` / `title: float` in `api-defs/agent/asyncapi.yml`. But the live service sends **numbers** — from a real session:

```
{"type":"AgentStartedSpeaking","tts_latency":0.105614436,"ttt_latency":0.645568172,"total_latency":0.75118354}
{"type":"LatencyReport","stt_latency":0.039999961853027344}
```

So I used a string-or-number deserializer on all ten fields rather than flipping the type to `String`, which parses whichever the service ends up sending. Serialization still writes numbers. Fixtures for both forms, plus an assertion that the two forms produce equal values and that a non-numeric string is still rejected.

Also worth knowing for your own model of the endpoint: `LatencyReport` arrives as many separate frames each carrying exactly one measurement, not one frame with seven fields.

### B2 — Flux TTS expressivity — pushing back
You quoted the spec correctly; the spec is wrong. Four live sessions against `wss://agent.deepgram.com/v1/agent/converse`, Settings frames identical except the JSON type of `agent.speak.provider.expressivity`:

- `"expressivity": "2"` → `Welcome`, then `{"type":"Error","description":"Error parsing client message. Check the agent.speak field against the API spec.","code":"UNPARSABLE_CLIENT_MESSAGE"}`, session over
- `"expressivity": 2` → `Welcome`, `SettingsApplied`, full turn completes

Reproduced with `"0"` and `0`. The spec's own description gives the game away — it says "Accepts the whole numbers -2 to 2" while typing the schema as a string enum. My guess is it was copied from `SpeakV2Expressivity`, which is a `/v2/speak` **query parameter**, where everything is a string on the wire.

So serialization stays an integer and I made deserialization accept both forms. You still get the two fixtures you asked for — exact `Settings` and `UpdateSpeak` wire JSON — they just pin the integer. If you want the string form shipped anyway, say so and I'll change it, but it fails against prod today.

### Spec bugs to raise upstream
`api-defs/agent/asyncapi.yml` is wrong in two places — expressivity typed as a string enum for a JSON body field, and latency typed as strings when the service sends numbers. Any SDK generating off that spec hits both.

No semver event: no public signature changed, `Serialize` output is unchanged, `Deserialize` only got more permissive, B3 only changed `Debug` output.

Gates green: fmt, clippy, `cargo test --all --all-features` (410 unit, 152 doc, 15 integration binaries), `cargo doc --workspace --all-features`, all with `-D warnings`.

GregHolmes (approved):

PR #169, opened 2026-08-27
feat(listen): opt-in connect diagnostics with per-phase timings

dg-coreylweathers:
## Summary

Adds an opt-in `connect-diagnostics` cargo feature for diagnosing connection latency on live transcription (`/v1/listen`) WebSocket connects.

- Configuring a sink via `WebsocketBuilder::diagnostics` makes the SDK establish the connection in four individually timed phases (DNS, TCP connect, TLS handshake, WebSocket upgrade) and emit one `diagnostics::ConnectRecord` per connect attempt.
- Records are delivered from a drop guard, so an attempt cancelled by a caller-side `tokio::time::timeout` still emits a record carrying the furthest phase reached and every completed phase timing.
- Failed upgrades capture the `dg-request-id` and `dg-error` response headers; a client-generated `attempt_id` correlates attempts that never got a server response.
- Records serialize to flat JSON for JSONL pipelines (`schema_version: 1`, additive-only; `ConnectRecord` is `#[non_exhaustive]` with a `sample()` constructor for sink tests).
- Without a sink — and always without the feature — the stock connect path runs unchanged. The phase-timed path reuses the same TLS provider, trust roots, and upgrade machinery as `tokio-tungstenite`'s `rustls-tls-webpki-roots` path, with a unit test asserting config equivalence.
- `WebsocketBuilder` now borrows the `Deepgram` client instead of the `Transcription` handle, so `dg.transcription().stream_request_with_options(...)` chains from a temporary (strictly more permissive; `cargo semver-checks` passes).
- Other WebSocket surfaces (Flux) are not covered yet.

Also rides along: fixes for two pre-existing clippy 1.98 `chunks_exact_to_as_chunks` warnings in the `text_to_speech_to_stream` example that would fail the Clippy job (`RUSTFLAGS: -D warnings`) on any PR run after the lint landed in stable.

## Test plan

- [x] `cargo test --all --all-features` — 71 unit tests (incl. 8 diagnostics tests), 5 localhost mock-server wire tests (success, missing/malformed request ID, rejected upgrade with header capture, stock path emits nothing), 139 doctests
- [x] Live e2e: `cargo test --features connect-diagnostics --test connect_diagnostics_e2e -- --ignored` — stock and diagnostic paths both connect to production `/v1/listen` with all four phase timings present; a 1ms-timeout cancellation still emits a record
- [x] `cargo check --all-targets --no-default-features --features=connect-diagnostics` (new CI matrix entry) and bare `--no-default-features`
- [x] `rustup run nightly cargo build --all-targets --all-features -Z minimal-versions` — new TLS crates compile at floor versions
- [x] `cargo fmt --check`, `cargo clippy --all-targets --all-features` with `-D warnings`, `cargo semver-checks check-release`

## Notes for reviewers

- Version is left at 0.10.0 with an `Unreleased` changelog heading; happy to bump to 0.11.0 in this PR if we'd rather follow the 0.10.0 (flux) precedent of bumping in the feature PR.


GregHolmes (changes requested):
Corey, I reproduced four blocking issues before this is merge-ready.

## [B1] Diagnostic URLs leak arbitrary query secrets

**Summary:** In `src/diagnostics.rs:97`, `DiagnosticsGuard::new` stores the complete request URL for JSONL logging. A local wire reproduction set a synthetic `customer_token` through `OptionsBuilder::query_params`; the exact value appeared in `ConnectRecord.url`. Signed callback URLs, keyterms, search/replace content, tags, and other arbitrary query values are exposed the same way.

**Expected:** Diagnostics telemetry must not persist credentials or customer-controlled text by default.

**Observed:** Every request query value is copied verbatim into the diagnostic record.

**Recommended fix:** Remove userinfo and query values from the recorded URL. If configuration context is needed, expose separately reviewed allowlisted non-sensitive fields, and add a regression test proving a synthetic secret is absent from serialized records.

## [B2] Enabling diagnostics changes stock TLS behavior

**Summary:** In `Cargo.toml:46-48`, direct `rustls` and `tokio-rustls` dependencies enable defaults that `tokio-tungstenite` explicitly disables. The resolved Cargo graphs confirm `connect-diagnostics` adds Rustls `logging` and `prefer-post-quantum`, while the listen-only graph does not. This affects the stock `connect_async` path even when no diagnostics sink is configured.

**Expected:** Enabling the Cargo feature without calling `.diagnostics(...)` must leave stock TLS negotiation and latency characteristics unchanged, and the timed path must use equivalent TLS settings.

**Observed:** Feature activation changes the unified Rustls configuration, contradicting the README and changelog's unchanged-path claim.

**Recommended fix:** Set `default-features = false` on the direct TLS dependencies and verify the listen-only and diagnostics feature graphs match. For downstream feature unification, use one explicit connector/configuration for both stock and timed paths.

## [B3] The diagnostics example produces invalid JSONL and exits successfully on failure

**Summary:** In `examples/transcription/websocket/connect_diagnostics.rs:40-80`, diagnostic JSON and human-readable failures both go to stderr, while connection, timeout, stream, and writer errors are swallowed. Running the example with a synthetic invalid key emitted prose and JSON on the same stream, then exited with status 0.

**Expected:** One documented stream must contain JSON records only, and operational failures must return a nonzero exit.

**Observed:** Failed runs corrupt the advertised JSONL stream and report success to automation.

**Recommended fix:** Emit diagnostic JSON to stdout, route human output to stderr, document redirection to a JSONL file, drain the writer, and propagate connection, timeout, stream, serialization, and writer-task failures.

## [B4] The ride-along Clippy edit breaks Rust 1.87 builds

**Summary:** In `examples/speak/rest/text_to_speech_to_stream.rs:166` and `:194`, `chunks_exact(2)` was replaced by `slice::as_chunks`, which stabilized in Rust 1.88. The official Rust 1.87 Docker image fails both changed lines with `E0658`.

**Expected:** This diagnostics PR must preserve existing compiler compatibility unless an MSRV increase is explicitly declared, tested, and communicated.

**Observed:** The existing TTS example no longer compiles on Rust 1.87.

**Recommended fix:** Restore `chunks_exact(2)` with a narrow `clippy::chunks_exact_to_as_chunks` allowance. If Rust 1.88 is the intended floor, separate that change, declare `package.rust-version`, add MSRV CI, and document it.

All four findings are covered by a disposable verifier and reproduce in one pass. The existing unit, integration, doc, feature-isolation, format, and current-toolchain Clippy checks otherwise pass.

GregHolmes (changes requested):
Thank you @dg-coreylweathers  I think there's still one small issue which is from B2: 

B2 Mixed downstream TLS features still break connector equivalence  

Summary: A valid consumer enabling tokio-tungstenite/native-tls makes stock connections use native TLS while diagnostics remains hardcoded to Rustls.  

Expected: Timed and untimed connections use the same provider and trust configuration.  

Observed: The disposable mixed-TLS consumer confirms both backends are enabled and the two code paths select differently.  

Recommended fix: Pass one explicit Rustls connector to both stock and diagnostic paths.

Resolved:
- B1 URL secret leakage
- B3 JSONL and exit behavior
- B4 Rust 1.87 compatibility

GregHolmes (dismissed):

GregHolmes (approved):

PR #170, opened 2026-08-31
feat(flux): add ForceEndTurn, TurnInfo trigger, and per-word timings

dg-coreylweathers:
## Summary

Phase 1 of full Flux v2 coverage: fills the three gaps between the crate's existing Flux STT support and the `schemas.listen.v2.yml` asyncapi spec.

- **`FluxHandle::force_end_turn()`** — sends `{"type":"ForceEndTurn"}` to end the current turn immediately on an external signal (push-to-talk release, DTMF tone, UI event, external VAD/endpointing). Sent while no turn is active, it is silently ignored — stated in rustdoc, the example, and the changelog.
- **`FluxResponse::TurnInfo.trigger: Option<TurnTrigger>`** — the cause of a turn ending, present on `EndOfTurn` events only: `Model` (native detection), `Manual` (a `ForceEndTurn` was sent), or `Timeout` (`eot_timeout_ms` elapsed). Open enum — unrecognized values deserialize as `Unknown`, per the spec's tolerance requirement.
- **`FluxWord.start` / `end`** — optional per-word timings in seconds.

Additive only; `TurnInfo` and `FluxWord` are `#[non_exhaustive]`, so external matchers/readers are unaffected.

## Testing

- 9 new serialization unit tests pin the exact wire JSON (ForceEndTurn/CloseStream formats, all three trigger values round-trip, unknown-trigger tolerance, optional word timings).
- `cargo build` + `cargo test --all-features` green on host and in Docker (`rust:1` + libasound2-dev): 68 unit + 3 integration + 136 doc tests. Clippy/fmt clean.
- **Live-verified** against `wss://api.deepgram.com/v2/listen`: ForceEndTurn sent mid-turn produced `EndOfTurn` with `trigger: manual` and the transcript-so-far; the next turn ended naturally with `trigger: model`. Also observed (and documented): ForceEndTurn during leading silence (no active turn) is silently ignored.
- New example: `flux_force_end_turn` (reads `DEEPGRAM_API_KEY`), which sends ForceEndTurn on the first non-empty `Update` so a turn is guaranteed active.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- `FluxHandle::force_end_turn()` sends the `ForceEndTurn` control message to end the current turn immediately on an external signal (push-to-talk release, DTMF tone, UI event, or an external endpointing stack). Note that `ForceEndTurn` is gated per deployment: where it is not enabled, the server replies with a fatal `UNPARSABLE_CLIENT_MESSAGE` error and closes the connection.
+- `FluxResponse::TurnInfo` now exposes `trigger` (`Option<TurnTrigger>`), the cause of a turn ending, reported on `EndOfTurn` events: `Model` (native end-of-turn detection), `Manual` (a `ForceEndTurn` was sent), or `Timeout` (`eot_timeout_ms` elapsed). `TurnTrigger` is an open enum — unrecognized values deserialize as `TurnTrigger::Unknown`.
+- `FluxWord` now exposes optional per-word `start` and `end` timings, in seconds.
+- New example: `flux_force_end_turn` under `examples/transcription/flux/`.
+
 ## [0.10.0](https://github.com/deepgram/deepgram-rust-sdk/compare/0.9.2...0.10.0)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `FluxHandle::force_end_turn()` sends the `ForceEndTurn` control message to end the current turn immediately on an external signal (push-to-talk release, DTMF tone, UI event, or an external endpointing stack). Note that `ForceEndTurn` is gated per deployment: where it is not enabled, the server replies with a fatal `UNPARSABLE_CLIENT_MESSAGE` error and closes the connection.
+- `FluxHandle::force_end_turn()` sends the `ForceEndTurn` control message to end the current turn immediately on an external signal (push-to-talk release, DTMF tone, UI event, or an external endpointing stack). Sent while no turn is active, it is silently ignored.
 - `FluxResponse::TurnInfo` now exposes `trigger` (`Option<TurnTrigger>`), the cause of a turn ending, reported on `EndOfTurn` events: `Model` (native end-of-turn detection), `Manual` (a `ForceEndTurn` was sent), or `Timeout` (`eot_timeout_ms` elapsed). `TurnTrigger` is an open enum — unrecognized values deserialize as `TurnTrigger::Unknown(String)`, preserving the original wire value so it survives re-serialization.
 - `FluxWord` now exposes optional per-word `start` and `end` timings, in seconds.
 - New example: `flux_force_end_turn` under `examples/transcription/flux/`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `FluxHandle::force_end_turn()` sends the `ForceEndTurn` control message to end the current turn immediately on an external signal (push-to-talk release, DTMF tone, UI event, or an external endpointing stack). Note that `ForceEndTurn` is gated per deployment: where it is not enabled, the server replies with a fatal `UNPARSABLE_CLIENT_MESSAGE` error and closes the connection.
-- `FluxResponse::TurnInfo` now exposes `trigger` (`Option<TurnTrigger>`), the cause of a turn ending, reported on `EndOfTurn` events: `Model` (native end-of-turn detection), `Manual` (a `ForceEndTurn` was sent), or `Timeout` (`eot_timeout_ms` elapsed). `TurnTrigger` is an open enum — unrecognized values deserialize as `TurnTrigger::Unknown`.
+- `FluxResponse::TurnInfo` now exposes `trigger` (`Option<TurnTrigger>`), the cause of a turn ending, reported on `EndOfTurn` events: `Model` (native end-of-turn detection), `Manual` (a `ForceEndTurn` was sent), or `Timeout` (`eot_timeout_ms` elapsed). `TurnTrigger` is an open enum — unrecognized values deserialize as `TurnTrigger::Unknown(String)`, preserving the original wire value so it survives re-serialization.
 - `FluxWord` now exposes optional per-word `start` and `end` timings, in seconds.
 - New example: `flux_force_end_turn` under `examples/transcription/flux/`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `FluxWord` now exposes optional per-word `start` and `end` timings, in seconds.
 - New example: `flux_force_end_turn` under `examples/transcription/flux/`.
 
+### Fixed
+
+- The Flux connection worker now ends on the first terminal transport error (forwarding it exactly once) and completes the WebSocket closing handshake promptly when the server initiates the close, instead of leaving the close acknowledgement queued until socket teardown. After the session ends — peer close or terminal error — `FluxHandle::send_data`, `configure`, and `force_end_turn` return an error instead of silently discarding the message.
+
 ## [0.10.0](https://github.com/deepgram/deepgram-rust-sdk/compare/0.9.2...0.10.0)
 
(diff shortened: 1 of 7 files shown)


GregHolmes (changes requested):
# Review: deepgram-rust-sdk #170, feat(flux): add ForceEndTurn, TurnInfo trigger, and per-word timings
Classification: mixed, code-led with a docs lens on the changelog and runnable example
Verdict: request-changes

## Intent

Fill three gaps in Flux `/v2/listen`: send `ForceEndTurn`, expose the `EndOfTurn` cause, and expose optional per-word timings. Existing Flux behavior and response compatibility must remain unchanged; unknown future trigger values must be tolerated. `ForceEndTurn` is deployment-gated and should fail visibly where unsupported. No external issue or launch brief is linked, so this intent is derived from the PR description, current AsyncAPI, and repository changes.

## Blocking

- [B1] `src/listen/flux.rs:516`, inbound response backpressure can prevent `ForceEndTurn` from reaching the server.
- [B2] `examples/transcription/flux/force_end_turn.rs:65`, the example sends a containerized WAV while declaring raw Linear16.
- [B3] `examples/transcription/flux/force_end_turn.rs:97`, fatal or unsuccessful demonstrations still return exit status 0.

**Title:** [B1] Response backpressure can starve `ForceEndTurn` (`src/listen/flux.rs:516`)

**Summary:** The single Flux worker prioritizes `ws_stream_recv` with `select_biased!` and awaits delivery into a bounded 256-item response channel at line 522. A deterministic mock server queued 600 valid `TurnInfo` responses; `force_end_turn()` returned `Ok(())`, but the server did not receive `ForceEndTurn` until the client drained all 600 responses. The turn can therefore end naturally or by timeout while the manual control remains only locally queued.

**Expected:** `ForceEndTurn` must remain deliverable after earlier outbound audio even when inbound updates are continuous or the response consumer is backpressured.

**Observed:** Inbound response forwarding blocks the same task responsible for writing controls, so the new core operation can be delayed indefinitely.

**Recommended fix:** Separate WebSocket reads and writes while retaining one FIFO outbound writer, with coordinated cancellation and first-error propagation. Add a mock-server regression test that fills the response channel and proves `ForceEndTurn` reaches the wire before responses are drained.

**Title:** [B2] The example declares WAV container bytes as raw PCM (`examples/transcription/flux/force_end_turn.rs:65`)

**Summary:** `bueller-mono.wav` is a RIFF/WAVE container, but the builder sets `encoding=linear16&sample_rate=44100`, which the AsyncAPI reserves for non-containerized audio. A local wire capture confirmed that the first binary frame starts with `RIFF` while the upgrade URL declares raw Linear16, so the WAV header is interpreted as audio. Production happened to transcribe the sample successfully, but the request remains contract-invalid and teaches developers the wrong integration pattern.

**Expected:** Containerized WAV bytes must be sent without `encoding` and `sample_rate`, or the example must parse the WAV and stream only its PCM payload.

**Observed:** The complete WAV, including its container header, is sent as raw Linear16.

**Recommended fix:** Minimally remove `.encoding(Encoding::Linear16)` and `.sample_rate(44_100)` so Deepgram detects the WAV container. Alternatively, decode the data chunk and keep the raw-audio parameters. Add a mock wire test asserting that RIFF bytes and raw-audio query parameters are never combined.

**Title:** [B3] Failed demonstrations exit successfully (`examples/transcription/flux/force_end_turn.rs:97`)

**Summary:** `FatalError` only prints and breaks, an early end of the response stream also breaks, and the example never requires an `EndOfTurn` with `trigger: Manual`. Unsupported deployments, connection loss after upgrade, or a race where the turn ends before the control arrives can therefore return `Ok(())` without demonstrating the advertised feature.

**Expected:** The example must return a nonzero exit for fatal or incomplete sessions, and success must require observing a manual end-of-turn result.

**Observed:** Every break path reaches the unconditional `Ok(())` at line 149, regardless of whether `ForceEndTurn` worked.

**Recommended fix:** Return an error immediately for `FatalError`, track whether `TurnTrigger::Manual` was observed, and return an error if the stream ends without it. Preserve the successful production path verified in this review.

## Should-fix

- [S1] `src/common/flux_response.rs:435`, unknown trigger strings are lost and reserialized as `"unknown"`.
- [S2] `src/listen/flux.rs:463`, `force_end_turn()` documentation does not distinguish local enqueue from server acceptance.
- [S3] `examples/transcription/flux/force_end_turn.rs:16`, the example contains two unresolved intra-doc links.
- [S4] `examples/transcription/flux/force_end_turn.rs:84`, post-close draining has no timeout and can hang indefinitely.

**Title:** [S1] Unknown trigger values do not round-trip (`src/common/flux_response.rs:435`)

**Summary:** `#[serde(other)]` maps every future trigger string to the unit `TurnTrigger::Unknown`. A local round-trip reproduction deserialized `"future_trigger"` successfully but serialized it back as `"unknown"`, discarding the server-provided cause. This weakens telemetry and conflicts with the response type's existing exact round-trip behavior.

**Expected:** An open enum should tolerate and preserve the original unknown wire value.

**Observed:** All unknown values collapse into one unit variant and lose their text.

**Recommended fix:** Use `Unknown(String)` with custom scalar serialization/deserialization, then add a test asserting an unknown trigger survives an exact round trip.

**Title:** [S2] `force_end_turn()` overstates synchronous success (`src/listen/flux.rs:463`)

**Summary:** The method's `Result` confirms only that the message entered the local channel. Server acceptance arrives asynchronously as `TurnInfo`, unsupported deployments report `FatalError` through `receive()`, and calls with no active turn are silently ignored. The current rustdoc states the successful outcome without explaining those semantics.

**Expected:** Public documentation must tell callers how to distinguish local enqueue, server-confirmed manual completion, idle no-op, and deployment rejection.

**Observed:** A returned `Ok(())` can be read as confirmation that the turn ended manually.

**Recommended fix:** State that `Ok(())` means locally queued; callers must observe `EndOfTurn` with `trigger: Manual`, idle calls produce no response, and unsupported deployments surface `FatalError` from `receive()` before closure.

**Title:** [S3] Example intra-doc links do not resolve (`examples/transcription/flux/force_end_turn.rs:16`)

**Summary:** In an example crate, `crate::common` refers to the example itself rather than the `deepgram` library. Running `cargo rustdoc --example flux_force_end_turn --features listen -- -D rustdoc::broken_intra_doc_links` fails both the `FluxResponse::TurnInfo` and `TurnTrigger::Manual` links.

**Expected:** Public example documentation must build without broken links.

**Observed:** Both links are unresolved when rustdoc documents the example target.

**Recommended fix:** Use `deepgram::common::flux_response::...` paths or plain inline-code references, and add example rustdoc to the documentation gate if these links are intended to remain clickable.

**Title:** [S4] Shutdown draining has no deadline (`examples/transcription/flux/force_end_turn.rs:84`)

**Summary:** After the file ends, the timer branch is disabled and the loop waits only on `handle.receive()`. If the peer or network stalls after `CloseStream`, the example never exits.

**Expected:** A runnable network example should bound post-close draining and fail visibly when shutdown does not complete.

**Observed:** There is no timeout once `audio_done` becomes true.

**Recommended fix:** Start a short shutdown deadline after `close_stream()` and return an error if the response stream does not terminate before it expires.

## Nits

No additional nits. The transport and example correctness issues above should stay the focus.

## Verified (evidence)

- PASS: Fresh comparison against `origin/main` at `52711b7`; the local diff matches GitHub at five files and `+311/-1`.
- PASS: `ForceEndTurn` serializes exactly as `{"type":"ForceEndTurn"}`, matching `schemas.listen.v2.yml` and `operations/listen.v2.yml`.
- PASS: Known trigger names (`model`, `manual`, `timeout`), optional trigger placement, and optional `FluxWord.start`/`end` fields match the AsyncAPI schema.
- PASS: The added fields are semver-safe because `FluxResponse::TurnInfo` and `FluxWord` were already non-exhaustive; `TurnTrigger` is also non-exhaustive.
- PASS: Multiple or empty `StartOfTurn` events do not affect production code because this client forwards events without assembling turn state. The example waits for a non-empty `Update`.
- PASS against production: the 17.6-second sample produced `EndOfTurn` with `trigger: Manual` for the forced first turn and `trigger: Model` for later turns, then closed cleanly. The exported key was passed directly to the SDK and was not printed or inspected.
- PASS: `cargo test --all --all-features`: 68 unit tests passed, 2 ignored; 3 integration tests passed, 1 ignored; 136 doc tests passed.
- PASS: local Rust 1.94.1 Clippy with warnings denied, format, listen-only feature build, and diff check.
- REPRODUCED: `/var/folders/v9/6bdxf7bx7fj967wbg6x_jdrr0000gn/T/opencode/pr170-repro` proves B1, captures B2 on the wire, and proves S1 rewrites an unknown trigger to `"unknown"`.
- FAIL: example rustdoc with broken-link warnings denied, confirming S3.
- FAIL in GitHub CI on Rust 1.98: inherited `chunks_exact(2)` warnings at `examples/speak/rest/text_to_speech_to_stream.rs:166` and `:192`. The same failure is on `main`-based PRs and is not introduced by #170, but the required check must be green before merge.

## Needs human

- Corey/server team: confirm whether a public method that produces a fatal error on deployments where the gate is disabled is acceptable to ship now, or whether merge/release should wait for deployment coverage.
- Corey/Greg: decide where to clear the inherited Rust 1.98 Clippy failure. PR #169 attempts that fix but is itself under requested changes; #170 cannot merge while the required check remains red.

## Developer-facing messaging

The changelog correctly identifies the deployment gate and distinguishes `Manual`, `Model`, and `Timeout`. Keep that framing, but make the example and rustdoc require observable manual completion so developers do not treat local enqueue or a zero exit code as server confirmation.

GregHolmes (dismissed):
Approving with nits:

All seven original findings are resolved.

Title: S5 Terminal errors do not stop the worker  
Summary: Read/write errors are forwarded, but processing continues.  
Expected: First terminal error ends the worker.  
Observed: Later commands and duplicate errors remain possible.  
Recommended fix: Break after forwarding terminal errors.

Title: S6 Peer close acknowledgement is not flushed  
Summary: A local probe confirmed the server receives no close acknowledgement while the handle remains alive.  
Expected: Complete the WebSocket closing handshake promptly.  
Observed: The acknowledgement remains queued until socket teardown.  
Recommended fix: Flush after receiving Message::Close.

GregHolmes (dismissed):

GregHolmes (approved):

PR #171, opened 2026-08-31
feat(speak): add Flux TTS support (/v2/speak REST and websocket)

dg-coreylweathers:
## Summary

Phase 2 of full Flux v2 coverage (**stacked on #170** — merge that first, then retarget or let GitHub auto-retarget this to `main`): a new `speak::flux` module implementing Deepgram's Flux text-to-speech v2 API per `schemas.speak.v2.yml` / `channels/speak.v2.yml` (asyncapi) and `paths/speak.v2.yml` / `parameters.speak.v2.yml` (openapi).

- **Batch (REST)**: `Speak::flux_speak_to_file` / `flux_speak_to_stream` send `POST /v2/speak` with a `{"text"}` body; options cover `model` (required — Flux has no default), `encoding`, `sample_rate`, `speed`, `expressivity`, `mip_opt_out`, `tag`, and the REST-only `container`, `bit_rate`, `callback`, `callback_method`, `priority`.
- **Streaming (WebSocket)**: `Speak::flux_request(options).handle()` connects to the `/v2/speak` websocket; `FluxSpeakHandle` sends `Speak`, `Flush`, `Interrupt` (optional `time_ms` playback offset), `Configure` (speed), `Close`. `FluxSpeakResponse` surfaces binary `Audio` frames plus every JSON event in the spec: `Connected`, `SpeechStarted`, `Flushed`, `SpeechMetadata`, `SpeechInterrupted`, `SessionMetadata`, `ConfigureSuccess`/`ConfigureFailure`, `Warning`, `FatalError`, and a forward-compatible `Unknown`.
- The websocket rejects REST-only options up front with the new `DeepgramError::InvalidOptions` variant instead of a confusing server-side failure.
- Abnormal (non-1000) server close frames are forwarded to the consumer as `WebsocketClose` errors rather than silently ending the stream.
- The `speak` feature now enables the websocket deps (`tungstenite`, `tokio-tungstenite`); `WsError` is available under `speak` as well as `listen`. Purely additive — no change for default-feature users, and each feature still compiles in isolation.

## Testing

- 26 new unit tests pin the wire formats of all five client messages and the deserialization of every server event (including omit-vs-present optional fields and unknown-type tolerance).
- `cargo build` + `cargo test --all-features` green on host and in Docker (`rust:1` + libasound2-dev): 94 unit + 3 integration + 138 doc tests. Clippy/fmt clean.
- **Live-verified** against production: batch produced a valid MP3 (48 kbps, 24 kHz mono); the websocket ran the full documented lifecycle in order (`Connected` → `SpeechStarted` → `Flushed` → audio → `SpeechMetadata` → `SessionMetadata` → clean close), with the WAV duration byte-math matching `audio_duration_ms` exactly; `configure_speed(3.5)` returned `ConfigureFailure SPEED_OUT_OF_RANGE` with the spec's exact field/value shape; `configure_speed(1.05)` returned `ConfigureSuccess`; `interrupt(Some(500))` returned `SpeechInterrupted` with the correct `text_spoken`/`text_remaining` split and rebased session totals.
- New examples (both read `DEEPGRAM_API_KEY`): `flux_tts_batch` and `flux_tts_websocket` (collects raw linear16 and writes a playable WAV).

## Note for reviewers

Live testing surfaced a spec-example divergence (not an SDK issue): `Connected.model_name` returns the short voice name (`haley`) rather than the spec example's full model string (`flux-alexis-en`). Both fields are string-typed so nothing breaks; a deepgram-docs spec-example refresh is being routed separately.


GregHolmes (changes requested):
Corey, I reproduced three blocking issues before this is merge-ready.

## [B1] Control messages stall behind response backpressure

**Summary:** In `src/speak/flux/websocket.rs:323`, the single worker prioritizes inbound WebSocket messages and awaits writes to the bounded response channel. A deterministic mock server queued 600 audio responses; an `Interrupt` sent by the client did not reach the server until all 600 responses were drained.

**Expected:** `Interrupt`, `Configure`, and `Close` remain deliverable while audio is arriving and while the response consumer is backpressured.

**Observed:** Control messages can remain queued behind inbound audio, so barge-in continues speaking and graceful shutdown can stall.

**Recommended fix:** Run the WebSocket read and write halves independently with coordinated cancellation and first-error propagation. Add a regression test that fills the response channel and proves `Interrupt` reaches the mock server before responses are drained.

## [B2] Batch-only encodings bypass WebSocket validation

**Summary:** In `src/speak/flux/options.rs:269`, `rest_only_options_set()` checks REST-only fields but not the known REST-only encodings. A production handshake using `Encoding::Mp3` returned `WsError`, confirming that the SDK attempted the connection instead of rejecting it locally.

**Expected:** `Encoding::Mp3`, `Opus`, `Flac`, and `Aac` return `DeepgramError::InvalidOptions` before `connect_async` is called.

**Observed:** These encodings are added to the WebSocket URL and left for the server to reject.

**Recommended fix:** Reject those four known compressed encodings during transport validation, retain server validation for `CustomEncoding`, and add one test for each rejected variant.

## [B3] Fatal synthesis creates a successful WAV artifact

**Summary:** In `examples/speak/flux/websocket_synthesize.rs:117`, `FatalError` only breaks the receive loop. A mock fatal event following this control flow produced a 44-byte header-only WAV, printed the success path, and would return exit code 0.

**Expected:** Fatal, incomplete, or empty synthesis exits with an error and does not create an output artifact.

**Observed:** The example unconditionally writes the collected bytes and reports `Audio saved` after a fatal event or premature close.

**Recommended fix:** Return an error for `FatalError`, require terminal `SessionMetadata`, reject empty audio, and create the output file only after those checks pass.

The rest of the modeled OpenAPI and AsyncAPI surface, feature isolation, serialization, and WAV header bytes checked out. Local tests passed: 94 unit tests, 3 integration tests, and 138 doc tests. PR #171 is still stacked on #170, and both currently inherit the same Rust 1.98 Clippy failure in `examples/speak/rest/text_to_speech_to_stream.rs:166` and `:192`; that required check also needs to be cleared before merge.

GregHolmes (changes requested):
Sorry, one thing I believe we should resolve first:

Title: **B4** Peer close leaves a silently successful send path  
Summary: After server closure, receive() returns None, but speak() still returns Ok and its message is discarded. The close acknowledgement is also never flushed.  
Expected: Peer closure terminates the worker, acknowledges the close, and makes later sends fail.  
Observed: The worker drains commands until the handle is dropped.  
Recommended fix: Flush the close reply, drop the command receiver, and return immediately.

GregHolmes (dismissed):

GregHolmes (approved):

PR #175, opened 2026-09-09
feat(tls): explicit WebSocket TLS trust, rustls-tls-native-roots feature, Deepgram::tls_config

dg-coreylweathers:
## Why

Twilio reported that enabling the 0.10.1 `connect-diagnostics` feature breaks websocket connections behind their Zscaler proxy. Root cause: our crate has never enabled tokio-tungstenite's `rustls-tls-native-roots`. Twilio's build enables it through another crate, and tokio-tungstenite's *default* connector merges the OS certificate store when that feature is unified in. 0.10.0 used that default connector, so the Zscaler CA was trusted by accident. The explicit webpki-only connector 0.10.1 introduced (so the timed and untimed paths could not diverge) bypassed that merge. Our regression, not their environment.

Twilio asked for a native-roots feature or a way to supply the TLS connector directly. This PR does both, and fixes the class of bug rather than the instance.

## What

- **`rustls-tls-native-roots` cargo feature.** Also trusts the OS certificate store, merged **on top of** the bundled webpki roots (never instead, so containers with no OS store keep working). Named after the tokio-tungstenite and reqwest features it mirrors. For Twilio: one word in `Cargo.toml`, no code change.
- **`Deepgram::tls_config(impl Into<Arc<rustls::ClientConfig>>)`.** Supply your own rustls config once, on the client; every websocket it opens uses it verbatim. `rustls` is re-exported as `deepgram::rustls` so versions match.
- **`DeepgramError::UntrustedTlsCertificate { host, trust, source }`.** Returned instead of a bare `WsError` on an unknown issuer; the message names the remedy for the trust roots in effect.
- **One explicit connector for every websocket surface.** Live transcription, Flux STT, and Flux TTS now all connect through the client's rustls config (new `src/tls.rs`). Trust can no longer differ by surface or be changed by downstream feature unification. Flux users who relied on the unification accident need the feature too; the changelog says so plainly.
- **Diagnostics.** Records gain `tls_trust` and `tls_resumed` (schema version unchanged, additive). The TLS config is resolved before the DNS timer starts, so an OS-store read is never charged to `tls_handshake_ms`. Default config is built once per client, so TLS sessions resume across connections from the same client.
- **Deps.** `rustls`, `tokio-rustls`, `rustls-pki-types`, `webpki-roots` move from `connect-diagnostics` to `listen`/`speak` via an internal `__tls` feature (already in the graph through tokio-tungstenite; nothing new downloads). rustls floor raised to `^0.23.27` for `handshake_kind()`. `tracing` bounded to `<0.2`.

## From Lawrence's branch

The trust-merge design (webpki always present, OS roots additive, warn-and-continue on bad certs, mirroring upstream) and the `SSL_CERT_FILE` test fixture come from `feat/native-roots-tls-trust`, with co-author credit. Changed from that branch: a feature flag instead of a per-builder method, no process-wide cache or TTL (the config lives on the client instead), and the code moved out of the diagnostics module so every surface uses it.

## Tests and gates

- New `tests/common/mod.rs` (local TLS server), `tests/connect_tls_config_local.rs` (7 tests), `tests/connect_native_roots_local.rs` (5 tests): every surface, both connect paths, feature on and off. Includes the session-resumption assertion and the error-hint assertions.
- `cargo fmt`, `cargo clippy --all-targets --all-features`, 9 feature combinations with `-D warnings`, `cargo test --all-features` (112 unit, 25 local integration, 142 doctests), default and listen-only test runs, `RUSTDOCFLAGS=-D warnings cargo doc --all-features --no-deps`, 1.87 toolchain check, nightly minimal-versions job as CI runs it.
- `cargo semver-checks --baseline-version 0.10.1 --all-features --release-type patch`: 223 pass, 0 fail. Ships as **0.10.2**.
- CI matrix: three new feature-combination checks and a default-feature test step.

## Before tagging

Twilio (jfoster) verifies this branch behind Zscaler with a git dependency. Then a separate `chore: release 0.10.2` PR.


GregHolmes (changes requested):
# Review: deepgram-rust-sdk #175, feat(tls): explicit WebSocket TLS trust, rustls-tls-native-roots feature, Deepgram::tls_config
Verdict: request-changes

## Intent
Restore explicit native-root trust for TLS-inspecting proxies, make WebSocket TLS trust consistent across live Listen, Flux STT, and Flux TTS, and offer a client-wide custom rustls configuration. The release is planned as 0.10.2 after Twilio verifies the branch behind Zscaler. This re-review verified the post-review README fix; no earlier GitHub review or local review record was available to map.

## Blocking
- [B1] `src/lib.rs:426` and `README.md:120`, `tls_config` is documented as covering plaintext WebSockets.

**Title:** [B1] TLS configuration documentation falsely covers `ws://` connections

**Summary:** `Deepgram::tls_config` and the README say the supplied rustls configuration is used by every WebSocket connection. The SDK deliberately converts an `http` base URL to `ws://`, and the phase-timed connector explicitly takes its `MaybeTlsStream::Plain` path for that scheme; no TLS handshake or certificate verification occurs. The public self-hosted constructors document `http://` base URLs, so this is a realistic path, not only a theoretical transport distinction.

**Expected:** Documentation states that custom trust roots and TLS configuration apply only to `wss://` connections, and it warns that `ws://` is plaintext.

**Observed:** A developer can configure a private CA and reasonably believe it protects the documented self-hosted WebSocket connection even though the configuration is unused.

**Recommended fix:** Amend the `tls_config` docs and README to say "every WSS WebSocket connection" and add a short warning to use `https://` or `wss://` whenever credentials or private traffic are involved. Keep the current `ws://localhost` examples explicitly scoped to local testing.

## Should-fix
- [S1] `src/tls.rs:150`, native-root load errors are hidden but error hints report successful OS-store checking.

**Title:** [S1] Native-root failure produces a misleading certificate remedy

**Summary:** `rustls-native-certs` errors are written only to tracing and then discarded while the client continues with webpki roots. When an invalid or unreadable `SSL_CERT_FILE` produces no usable roots, the later `WebpkiAndNative` error says the OS store was checked and suggests pointing `SSL_CERT_FILE` at the CA, even though that attempted source may be the failed input.

**Expected:** Certificate errors distinguish successfully loaded native roots from a failed native-root load and offer a remedy that matches the actual trust store.

**Observed:** The customer receives an actionable-looking but false diagnosis unless they also happen to capture the tracing warning.

**Recommended fix:** Preserve native-root load status in the TLS settings and use it in `UntrustedTlsCertificate`; either surface the load failure directly or change the hint to state that native roots could not be loaded and recommend a valid `SSL_CERT_FILE` or `tls_config`.

- [S2] `src/diagnostics.rs:129`, plaintext diagnostic records claim certificate trust information.

**Title:** [S2] `tls_trust` is reported for connections without TLS

**Summary:** `ConnectRecord::tls_trust` is documented as the roots used to verify a server certificate and is always populated. The diagnostics connector records the configured trust mode even when the URL is `ws://` and takes the plain TCP path without a certificate.

**Expected:** A plaintext connection does not report TLS verification state as though a certificate was checked.

**Observed:** Consumers can read `tls_trust: webpki` for a `ws://` attempt with no TLS phase.

**Recommended fix:** Make `tls_trust` optional and omit it for `ws://`, or revise its documentation to say it is the configured trust mode and require callers to use `tls_handshake_ms` to determine whether TLS occurred. The optional field is clearer because this is an unreleased additive diagnostic field.

## Nits
- None.

## Verified (evidence)
- The follow-up README change correctly wraps DER bytes in `rustls::pki_types::CertificateDer` before calling `RootCertStore::add`; the doc test compiles.
- Client TLS routing is correct on all claimed surfaces: live Listen, Flux STT, Flux TTS, and the diagnostics path pass the client-owned connector/configuration.
- Default webpki roots, additive native roots, custom configurations, UnknownIssuer mapping, and session resumption are covered by the local TLS integration tests.
- Local Rust 1.94.1 passes `cargo test --all-features --no-fail-fast`, clippy, fmt, docs, and the reviewed feature combinations. CI reports all checks, including SemVer, passing.
- The branch merges cleanly with current `main`.

## Needs human
- Twilio's Zscaler validation remains the stated pre-release gate. The local fixture proves that `SSL_CERT_FILE` roots are merged and routed on every claimed WebSocket surface, but it cannot verify Twilio's proxy and deployment configuration. The release owner should hold the 0.10.2 tag until jfoster confirms the git-dependency test.

## Developer-facing messaging
- The changelog accurately explains the 0.10.1 regression and the additive native-root feature. It should make the same WSS-only distinction as [B1], especially because it mentions self-hosted deployments and credentials.

GregHolmes (approved):

PR #179, opened 2026-09-15
fix(listen)!: unblock and deliver the streaming terminal transport error

dg-coreylweathers:
## Proposed changes

The `/v1/listen` and `/v2/listen` streaming workers forwarded inbound responses and terminal errors with a blocking send, so a full response channel parked the worker mid-forward: it stopped writing the caller's audio, the caller filled the bounded command channel and parked in `send_data`, and the session hung. On `/v1/listen` this happened on a **perfectly healthy socket with no transport failure involved at all** — the inbound forward was never gated. Present on `/v1/listen` since 0.6.0 and on `/v2/listen` since 0.8.0.

Three changes:

- Inbound reads reserve capacity with `poll_ready` before forwarding, so backpressure reaches the socket instead of the worker — the behavior `/v2/listen` has had since 0.10.1.
- Terminal errors forward through a `response_tx.clone()` that owns a guaranteed channel slot (a `futures` mpsc channel's capacity is `buffer + num_senders`), so the error neither blocks nor gets dropped. Plain `try_send` was measured dropping it: 257 queued responses, 0 errors delivered.
- The first failed write ends the session, so one broken transport produces exactly one error instead of one per subsequent write plus one from cleanup.

The Flux text-to-speech WebSocket (`/v2/speak`) has the same defect; its fix ships on the Phase 3 branch (#166), so this PR does not claim it.

## Types of changes

The worker no longer drains its command channel after the loop, so sends on an ended `/v1/listen` session return `Err` where 0.11.0 returned `Ok(())` indefinitely. Per `AGENTS.md`, a behavior change that makes working code fail until the consumer changes something is breaking in 0.x, so the release carrying this needs a minor bump and the `**BREAKING**` line is in the changelog. `cargo semver-checks` passes — no signature moved — which is exactly why it is not the evidence here.

## Checklist

- [x] I have read the CONTRIBUTING.md doc

## Further comments

Each regression test was verified by reverting the fix and watching it fail, because a test that only asserts "does not hang" passes with the error dropped:

| Revert | Result |
| --- | --- |
| Full source revert | both regression tests fail, `Elapsed(())` at 31.03 s total |
| Gating kept, terminal forward back to blocking | write-error test fails, `Elapsed(())` at 16.02 s |
| Gating kept, `try_send` instead of the sender clone | `left: 0  right: 1` — the error was dropped |
| Command drain restored | send on an ended session succeeds forever instead of erroring |

The two causes mask each other: with only the error-forward fix, the original reproduction still hangs. The delivery test is coordinated on two oneshots and asserts `responses == 257`, so a run that failed to fill the channel fails loudly rather than passing vacuously.

**Known defect left for its own branch:** with `keep_alive` off (the default), the elapsed 3-second timer branch parks the worker in `pending::<()>()` permanently — a response arriving 5 s into an idle session never reaches the consumer. Pre-existing, different blast radius.


dg-coreylweathers:
## Pre-review before handing this to Greg: approve-with-nits, no blockers

### What this PR does

When your code stops reading `receive()` for a moment, the SDK's background worker used to freeze. It filled a 256-item queue, then sat waiting for room — which meant it also stopped sending your audio, and your next `send_data` waited forever. On `/v1/listen` this needed nothing to go wrong with the network at all. This PR makes the worker check for queue space *before* reading from the socket, sends any fatal transport error through a reserved slot that is always available, and ends the session on the first failed write. It also stops the worker accepting audio after the session is over, which is the breaking change in the changelog.

### What I checked

- **Do the new tests actually catch the bug?** Yes. I reverted both source files to `main`, kept the tests, and ran them: 3 of the 4 new tests fail. The fourth is a control that should pass either way. The revert table in the PR body holds up.
- **Is the fatal error really delivered exactly once?** Yes. Every path that ends the worker sets `is_open = false` at or before the forward, and the cleanup write is guarded by `if is_open`, so the one reserved slot is spent at most once. The Flux `CloseStream | None` arm sets the flag before its send, so it cannot double-report.
- **Does it work against the real API?** Yes. Ran it in Docker against `wss://api.deepgram.com/v1/listen` with a real key: connected, streamed 3.1 MB, and a `send_data` after the session ended returned an error instead of hanging.
- **Did the live run prove the `Ok` → `Err` change?** No, and I want to be straight about that. `main` returned an error there too, because a live server closes with a close frame, and that path drops the command channel on both versions. The changelog's "returned `Ok(())` indefinitely" case is the other exit path, which the mock-server test does cover and which does fail on reverted source.
- **Is the 3-second `pending()` park really pre-existing?** Yes. Same probe times out identically on this branch and on `main` at 0.11.0. This PR neither causes nor worsens it.
- **Written semver policy?** Met. AGENTS.md wants a `**BREAKING**` line leading with what breaks and what to do; CHANGELOG.md:12 has one.

### Two should-fixes, neither blocking

**[S1] The new error reads like an SDK bug.** A send after the session has ended is now expected and documented — but it surfaces as `DeepgramError::InternalClientError` (`src/listen/websocket.rs:935`, `:968`), which prints:

```
an unepected error occurred in the deepgram client: send failed because receiver is gone
```

That names an internal channel, reads as a defect in the SDK rather than a thing the developer did, and misspells "unexpected" (`src/lib.rs:263`). This PR is exactly what makes that string common, since it is the surface of the headline breaking change. Worth a dedicated variant saying the session has ended and to stop sending or reconnect — and the typo fix while you are in there.

**[S2] The known-defect note understates its reach.** The PR body describes the parked-worker defect as "a response arriving 5 s into an idle session never reaches the consumer", which reads as an idle edge case. It is broader: after `close_stream()`, `is_open` is false, so `keep_alive` stops keeping the worker alive and the 3-second timer parks it. I measured all three cases:

| keep_alive | session state | late response |
| --- | --- | --- |
| on | open | delivered |
| off (default) | open | never arrives |
| on | after `close_stream()` | never arrives |

So anyone following the documented shutdown (`close_stream()`, then drain until `None`) hangs whenever the server takes more than ~3 s to finalize. Pre-existing and identical on `main`, so it stays out of scope here — but it is worth restating in those terms so the follow-up branch is scoped right.

### One thing for the release, not for this PR

Nothing will tell you this branch is breaking. I ran `cargo semver-checks check-release` the way CI does against sibling PR #181, which changes a public field's type: *"196 checks: 196 pass, no semver update required"*. It has no lint for a changed field type, and with the version unchanged it compares 0.11.0 to 0.11.0. #179, #180 and #181 are all breaking at 0.11.0 with a green SemVer check, and #164–#167 already claim 0.12.0. The release PR has to pick 0.12.0 deliberately — same class of miss as the 0.10.2 → 0.11.0 relabel.

Full record with evidence in `REVIEW-pr179.md` (local, git-excluded).

GregHolmes (approved):

PR #181, opened 2026-09-15
fix(manage)!: decode usage requests that omit the audio meter group

dg-coreylweathers:
## Proposed changes

`dg.usage().list_requests()` could not decode a normal production response. The request reached the right URL and the API returned 200, then deserialization failed with `missing field "duration"` — so the method was unusable against the live API.

Four fields on `manage::usage::response::Details` become `Option`: `duration`, `total_audio`, `channels`, `streams`. The reference marks them required; in practice they arrive as an all-or-nothing block and are omitted entirely for `/v1/read`, `/v1/speak`, `/v2/speak`, `/v1/agent/converse`, and for `/v1/listen` and `/v2/listen` records that metered no audio.

Nothing else was over-constrained: the four `#[allow(missing_docs)]` markers on them are replaced with doc comments stating when the value is absent, and no type corrections were needed.

## Types of changes

Code that read `record.duration` as `f64` (or `channels`/`streams` as `usize`) no longer compiles — `error[E0369]: cannot multiply Option<f64> by f64` against 0.11.0-era code. Match on the `Option` or use `unwrap_or_default()`. Worth knowing: `cargo semver-checks` has **no lint for a public field's type changing**, so it reports "no semver update required" and CI would have waved this through. The `**BREAKING**` changelog line is the only guard.

## Checklist

- [x] I have read the CONTRIBUTING.md doc

## Further comments

Rather than trust serde's first error, field presence was diffed across **3,943 real records** — one project's full history, 40 pages — plus the `status=succeeded` and `status=failed` filtered pages. The four fields appear on 1,485 of 3,787 detail objects (39%), and an independent scan of 1,200 records found them strictly all-or-nothing: 693 with all four, 481 with none, **zero partial**. Every other required field was present in 100% of records, so the relaxation is exactly the right width.

Before: `list_requests` errored on the default page and on `status=succeeded`, and `get_request` errored on any record without `duration`. After: all 3,943 decode, 0 failures. Nine fixture tests are built from real captured records — audio-metering, a no-metering `/v1/read` record (the repro), a zero-audio Flux speech-to-text stream asserting `Some(0.0)` is distinct from `None`, a rejected 400 with no `details`, `/v1/speak` with `tts_details`, Voice Agent with `sts_details`, a `"response": null` record, an async record whose `callback` carries only `attempts`, and a mixed three-record page. Project, key, and model identifiers are replaced with synthetic values; presence and absence are untouched.

Findings for the docs team, all live-verified:

- `duration`, `total_audio`, `channels`, `streams` are documented as always returned. They are not.
- `limit` maximum is **100**, not the documented 1000; `limit=500` and `limit=1000` both return 400 `INVALID_QUERY_PARAMETER`, and the message says "maximum allowed number of pages" when the constraint is page size.
- Undocumented response fields the SDK currently drops: `project_uuid` (3943/3943), `response.token_details` (3787), `response.deployment` (3787), `response.tts_details` (1068), `response.sts_details` (831), `response.method` (152, only on failed records where `details` is absent), `details.tier` (1485), `details.metadata` (1485). `usage` results also carry `tokens {in,out}` and `tts {characters,requests}`; `usage/fields` carries a top-level `accessors` array and a per-model `multilingual` flag.
- `details.config` is `{}` in all 3,787 records, though the SDK models 14 fields inside it.
- `/v1/agent/chatty` appears as a request `path` but is not in the documented `endpoint` filter enum.

**Adjacent gap left for its own PR:** `list_requests` cannot paginate — its options carry only `start`, `end`, `limit`, `status`, while the reference also documents `page`, `accessor`, `request_id`, `deployment`, `endpoint`, and `method`. With `limit` capped at 100 and no `page`, a caller cannot reach past the first 100 records. Additive.


GregHolmes (changes requested):
### [B1] `None` is documented as no audio metered when Voice Agent records show metered audio

**Summary:** The Voice Agent fixture includes `sts_details.duration: 0.071` while all four legacy `response.details` meter fields are absent. The new rustdoc and changelog say `None` means no audio was metered and recommend `unwrap_or(0.0)`, which can convert real Voice Agent audio usage to zero.

**Expected:** `None` must mean only that the legacy `response.details` meter group was omitted. It must not claim that the request incurred no audio-related usage.

**Observed:** `src/manage/usage/response.rs:83-106` and `CHANGELOG.md:12` make the no-audio claim. The included fixture at `src/manage/usage/response.rs:465-501` disproves it: `sts_details.duration` is `0.071` while the four legacy fields are absent.

**Recommended fix:** Reword the four public field comments to describe the absent legacy group without inferring metering semantics. Remove the changelog's `unwrap_or(0.0)` recommendation and tell callers to handle `None` deliberately. Modeling endpoint-specific meters such as `sts_details` can remain a separate additive change.

dg-coreylweathers:
## B1 fixed — and the live API makes the case stronger than the fixture did

Thanks, this was a real trap. Fixed in `b206f753` (documentation only, no signature or behaviour change).

**What changed**

- The four `Details` meter-field comments now describe only the absent legacy `response.details` group. `duration` carries the full explanation — an absent group does not mean the request metered no audio, because Voice Agent records report duration in a separate `sts_details` object this type does not model — and the other three point at it.
- The changelog `unwrap_or(0.0)` recommendation is gone, replaced with "decide deliberately what an absent value means for your accounting", plus an explicit note that reading `None` as zero can silently drop real usage.
- The `### Fixed` bullet had the same bad inference in its first clause ("sends ... only for requests that metered audio"). Reworded to state only which paths omit the group.

**Live verification**

Pulled 100 live usage records over a June-to-September window. 77 have the legacy meter group absent, across `/v1/read` (69), `/v2/speak` (4), `/v1/speak` (2) and `/v1/agent/converse` (2). Both Voice Agent records metered substantial audio while omitting the group:

| path | `sts_details.duration` | `details.usd` |
| --- | --- | --- |
| `/v1/agent/converse` | 358.018 s | 0.44752 |
| `/v1/agent/converse` | 304.493 s | 0.38062 |

So the recommendation I had written would have converted roughly six minutes of billed audio to zero on those two records alone. The in-repo fixture only showed 0.071 s, which understated it.

Modelling `sts_details` stays out of scope here, as you suggested — worth a separate additive change.

**Gates** (local, `RUSTFLAGS=-D warnings RUSTDOCFLAGS=-D warnings`): `fmt --check` clean, `clippy --all-targets --all-features` clean, `test --all --all-features` 126 unit + 142 doctests + all `*_local.rs` passing, `doc --workspace --all-features --no-deps` clean.

**Merge conflicts**: trial-merged this branch against every other open PR (#178, #179, #180, #182, and the #164-#167 stack). The only conflicted path in any pair is `CHANGELOG.md`, from parallel inserts into `## [Unreleased]` — no source-level conflicts. Whichever lands first, the rest need a one-hunk changelog touch-up. Clean against `main` as it stands.

dg-coreylweathers:
Pushed. B1 was already addressed by the earlier commit, but I found the same inference still sitting in two other places and fixed those, then rebuilt the whole claim on live data.

**The residual instance.** Both the `duration` rustdoc and the changelog still listed, among the cases where the API omits the meter group, "and for `/v1/listen` and `/v2/listen` records that carried no audio." That is the same absent ⇒ no audio reading you asked me to drop, and this PR's own fixture contradicts it — `LISTEN_STREAMING_ZERO_AUDIO` has the group **present with zeroes**, and `deserialize_zero_audio_is_some_not_none` asserts `Some(0.0)`.

**Then I checked it against real records instead of the spec.** 4,000 `list_requests` records, prod:

| endpoint | meter group |
|---|---|
| `/v1/listen` | always present when `details` is (1,392 records, 659 of them zeroed) |
| `/v1/read`, `/v1/speak`, `/v2/speak`, `/v1/agent/converse` | always omitted |
| `/v2/listen` | **present on 207, omitted on 13** — the omissions are all sessions that ran no model |
| any endpoint, 4xx/5xx | `details` itself absent |

So the split is per record, not per endpoint, and my first pass at rewording this got it wrong in the other direction before I had the data. The docs now say that, name the sibling fields each endpoint meters into (`sts_details`, `tts_details`, `token_details`), and there is a new `/v2/listen` fixture for the omitted shape so both `/v2/listen` shapes are covered.

Also merged `main` in — this had gone conflicting behind #178/#179. Only `CHANGELOG.md` conflicted; every bullet from both sides is kept.

Gates green: fmt, clippy, `cargo test --all --all-features`, `cargo doc --workspace --all-features`, all with `-D warnings`.

GregHolmes (approved):

GregHolmes:
Nit: `src/manage/usage/response.rs:326-328` still says the legacy meter group appears only for requests that "actually metered audio." This PR’s own `LISTEN_STREAMING_ZERO_AUDIO` fixture proves a zero-audio streaming request can carry the group with zeroes.

Please change the comment to: `A /v1/listen record with the legacy audio-metering group present.`

Non-blocking; the public rustdoc and changelog now correctly avoid inferring usage from the group’s presence or absence.

PR #182, opened 2026-09-18
fix(skills): use folded block scalar so SKILL.md frontmatter parses

dg-coreylweathers:
## Problem

`.agents/skills/deepgram-rust-voice-agent/SKILL.md` has YAML frontmatter that **does not parse**. The `skills` CLI therefore skips the skill, and this repo installs **6 of its 7 skills**.

The `description:` value was an unquoted plain scalar containing a colon-space sequence:

```
description: Use when a user asks for Deepgram Voice Agent support from Rust. Route honestly: this crate does not currently expose the Agent WebSocket API, ...
```

In YAML, the `honestly: this` inside a plain scalar reads as the start of a nested mapping, so the parse aborts. Exact error from `yaml@2.8.2` (the parser the `skills` CLI itself uses):

```
Nested mappings are not allowed in compact mappings at line 2, column 14:

description: Use when a user asks for Deepgram Voice Agent support from Rust. R…
             ^
```

The offending substring is `Route honestly: this crate does not currently expose ...`.

## Why this went unnoticed

The installer treats an unparseable skill as **non-fatal**: it emits one dim `⚠ Skipped` warning, drops the skill, reports a lower count, and **still exits 0**. Nothing fails, so the only visible symptom is a skill count that is quietly one short — and in CI, where only the exit code is checked, there is no symptom at all.

Note also that `deepgram/skills`'s `.claude-plugin/marketplace.json` lists **all 7** skills for this repo, including `.agents/skills/deepgram-rust-voice-agent`. So the published manifest and installable reality currently disagree.

## Proof (before / after)

Run in a clean `node:22-alpine` container, same command both times:

**Before** (`main`):

```
◇  Source: /repo
⚠ Skipped /repo/.agents/skills/deepgram-rust-voice-agent/SKILL.md — YAML parse error: Nested mappings are not allowed in compact mappings at line 2, column 14:
◇  Found 6 skills
EXIT=0
```

**After** (this branch):

```
◇  Source: /repo
◇  Found 7 skills
EXIT=0
```

Confirmed against the published repo too — `npx -y skills@latest add deepgram/deepgram-rust-sdk --list` reports `Found 6 skills` and exits 0, and a real `--yes` install writes only 6 skill directories:

```
INSTALL_EXIT=0
deepgram-rust-audio-intelligence
deepgram-rust-conversational-stt
deepgram-rust-management-api
deepgram-rust-speech-to-text
deepgram-rust-text-intelligence
deepgram-rust-text-to-speech
```

(`deepgram-rust-voice-agent` absent.)

## Fix

Convert `description` to a `>` folded block scalar, matching the house style already used by `deepgram/skills`'s capability skills (`skills/speech-to-text/SKILL.md` et al.), which is why those survive colons.

**The description text is unchanged.** Verified by re-parsing and comparing to the original string — identical apart from the single trailing newline a `>` folded scalar adds. No rewording or shortening, since these descriptions drive skill matching.

## Verification

A validator was run over the frontmatter of every `.agents/skills/*/SKILL.md` in all six public Deepgram SDK repos (42 skills) using `yaml@2.8.2`:

- Before: 2 failures repo-wide — this file and `deepgram-js-sdk`'s `deepgram-js-text-intelligence` (broken by `` `body: { text }` ``, fixed in deepgram/deepgram-js-sdk#558).
- After: **42/42 parse**, and every `name` still matches its directory name.

`deepgram-python-sdk`, `deepgram-java-sdk`, `deepgram-go-sdk` and `deepgram-dotnet-sdk` are all clean — no action needed there.

Leaving this for the Rust SDK owners to review and merge.


dg-coreylweathers:
## Independently verified, and it is the only manifest affected

Parsed the YAML frontmatter of all seven `SKILL.md` files on `main`. Six load fine; this one is the single failure:

```
OK    deepgram-rust-audio-intelligence    keys: [name, description]
OK    deepgram-rust-conversational-stt    keys: [name, description]
OK    deepgram-rust-management-api        keys: [name, description]
OK    deepgram-rust-speech-to-text        keys: [name, description]
OK    deepgram-rust-text-intelligence     keys: [name, description]
OK    deepgram-rust-text-to-speech        keys: [name, description]
FAIL  deepgram-rust-voice-agent           mapping values are not allowed here
```

The cause is the `: ` in `Route honestly: this crate does not currently expose...`. In an unquoted YAML scalar that reads as a nested mapping, so the parse fails on the *whole* frontmatter block — an agent loading this skill gets neither `name` nor `description`, not merely a truncated one. The folded block scalar here fixes it, and the rendered description is unchanged.

So the fix is correct, complete, and minimal as it stands.

## One scheduling note for whoever merges

#167 rewrites this same file end to end (it replaces the "not currently exposed" description, since that PR ships the Agent client). The two therefore conflict on this one path:

- **If this lands first** — #167 hits a one-file conflict, resolved by taking #167's version wholesale. Its replacement description contains no colon, so it parses cleanly; I checked.
- **If #167 lands first** — this PR becomes moot and should just be closed.

Worth landing this one now regardless: #167 sits at the top of a four-deep stack still in review, and until it lands `main` ships a skill manifest no agent can read.

GregHolmes (approved):

